PR #28, opened 2022-06-03
Paginate results implementation for PE & CE (LEAN-582)

asouqi:
As we are going to use pagination in the dashboard, we will not request all the submissions, so we will utilize `submissionId` to get both `projectID` & `manuscriptID` from [documentId](https://github.com/atypon/leanworkflow-graphql/blob/master/lean-workflow.graphqls#L115).

### DIFF
--- a/src/EditorApp.tsx
+++ b/src/EditorApp.tsx
@@ -60,7 +60,7 @@ const EditorApp: React.FC<Props> = ({ submissionId }) => {
     if (!data?.submission?.documentId) {
       return
     }
-    const [projectID, manuscriptID] = data?.submission?.documentId.split('#')
+    const [projectID, manuscriptID] = data.submission.documentId.split('#')
 
     // implement remount for the store if component is retriggered
     const basicSource = new BasicSource(
--- a/src/EditorApp.tsx
+++ b/src/EditorApp.tsx
@@ -62,6 +62,8 @@ const EditorApp: React.FC<Props> = ({
       submissionId,
       projectID,
       manuscriptID,
+      submission,
+      person,
       userID || ''
     )
     const couchSource = new CouchSource()
@@ -83,10 +85,7 @@ const EditorApp: React.FC<Props> = ({
         <NotificationProvider>
           <Page>
             <Wrapper>
-              <ManuscriptPageContainer
-                submission={submission}
-                person={person}
-              />
+              <ManuscriptPageContainer />
             </Wrapper>
           </Page>
         </NotificationProvider>
--- a/src/EditorApp.tsx
+++ b/src/EditorApp.tsx
@@ -9,6 +9,7 @@
  *
  * All portions of the code written by Atypon Systems LLC are Copyright (c) 2019 Atypon Systems LLC. All Rights Reserved.
  */
+import { ApolloError } from 'apollo-client'
 import React, { useEffect, useState } from 'react'
 import { hot } from 'react-hot-loader'
 import { BrowserRouter as Router } from 'react-router-dom'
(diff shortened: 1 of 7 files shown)


mbartenev-atypon:
I don't think that we should involve editor in pagination implementation for several reasons:
1. Article editor should be as least aware of lw-store as possible, as it may be potentially reused somewhere else. Yes, currently it won't work without the lw-store at all, but we shouldn't tighten them together even more when is not absolutely necessary.
2. Making article-editor solve this specific state management problem of its parental component (the dashboard) won't solve other state management problems of its parental component. However, if we will fix the state management problem of the editor as an integral problem, we won't have to involve the editor at all (and any other sub-components for that matter).
3. My understanding that the problem that we are trying to solve here is that the context in the store will not contain any info about the submission before editor is rendered. Hence, no data to be passed into. The same problem will occur for the preview page too and if we follow this approach, the preview page will require a different solution. If, however, we will resolve that problem on state management level, we won't have to apply an ad-hoc solution for each case, but will have a universal and well maintainable approach.
4. No chance of potentially caching this data on FE. Article editor will always request data, that are already in the memory of dashboard, but using a different apollo instance.

Possible solutions:
1. Do a useQuery(<submission>) on each page where it is required in the dashboard and make sure that apollo-cache effectively manages these requests along with paginated queries.
2. Decorate useQuery in GenericContext and when a component (EditorPage in this case ) asks for a submission, check if it is already in the store before querying server. That's more of a backup plan if apollo-cache won't work out for some reason.

asouqi:
> I don't think that we should involve editor in pagination implementation for several reasons:

pagination is not involved with the editor, and please clarify how it's involved with the editor?

>1. Article editor should be as least aware of lw-store as possible, as it may be potentially reused somewhere else. Yes, currently it won't work without the lw-store at all, but we shouldn't tighten them together even more when is not absolutely necessary.
>2. Making article-editor solve this specific state management problem of its parental component (the dashboard) won't solve other state management problems of its parental component. However, if we will fix the state management problem of the editor as an integral problem, we won't have to involve the editor at all (and any other sub-components for that matter).

editor tightly coupled on lw-store data for everything files submission data.

 `we won't have to involve the editor at all (and any other sub-components for that matter).`  
yes which is make sense

>3. My understanding that the problem that we are trying to solve here is that the context in the store will not contain any info about the submission before editor is rendered. Hence, no data to be passed into. The same problem will occur for the preview page too and if we follow this approach, the preview page will require a different solution. If, however, we will resolve that problem on state management level, we won't have to apply an ad-hoc solution for each case, but will have a universal and well maintainable approach.

preview page will work as we use submissionId to get the data when need for that page.

>4. No chance of potentially caching this data on FE. Article editor will always request data, that are already in the memory of dashboard, but using a different apollo instance.

Will it's not in the memory always, you go to the editor link directly and submission couldn't be found.

>Possible solutions:
>
>Do a useQuery() on each page where it is required in the dashboard and make sure that apollo-cache effectively manages these requests along with paginated queries.

why should I request each page for something we don't need!

>Decorate useQuery in GenericContext and when a component (EditorPage in this case ) asks for a submission, check if it is already in the store before querying server. That's more of a backup plan if apollo-cache won't work out for some reason.

the back end provides for us a query to get a submission by it's id, I'm not sure why we should add this complexity to the front end, the front end should just take the data and render it


if the plan is to make this editor separate on lw-store which is impossible but anyway the only option for me will be to get the `documentId` in the editor page in dashboard like the preview page and pass submission data to the `article-editor` component,

mbartenev-atypon:
> pagination is not involved with the editor, and please clarify how it's involved with the editor?
This change is in relation to pagination, judging by the title of this pull request.

> editor tightly coupled on lw-store data for everything files submission data.
Yes, but it shouldn't be. As I said while we don't pursue to loosen the coupling at this point, we shouldn't tighten it either. This is not only because we may use it elsewhere in the future but also because it will become increasingly harder to work with and maintain if we will tighten those components together.

> preview page will work as we use submissionId to get the data when need for that page.
Editor page shouldn't be any different.

> Will it's not in the memory always, you go to the editor link directly and submission couldn't be found.
Yes in this case, but if you browse into the editor page through submission -> inspector -> editor, the data will be there. Or if you will close the editor page and get back to the preview page or submission page and then go back to the editor, data will be queried again in this implementation, as the editor is destroyed every time.

> why should I request each page for something we don't need!
Hmm. I probably didn't make myself clear, but this is exactly what I mean. We should request on the data we need on those page (I mean webpages) on which we are at one particular moment. But if we queried something somewhere, it shouldn't be queried again until app is not destroyed.

> the back end provides for us a query to get a submission by it's id, I'm not sure why we should add this complexity to the front end, the front end should just take the data and render it

The frontend - in this case dashboard - needs a certain level of complexity, which is absent now and which will ensure state management. The way I see it - here we are trying to work around insufficient state management of lw-dashboard by making article-editor provide data for itself.
 
> if the plan is to make this editor separate on lw-store which is impossible but anyway the only option for me will be to get the documentId in the editor page in dashboard like the preview page and pass submission data to the article-editor component,

yes, this is kinda what I suggest

mbartenev-atypon (approved):

PR #60, opened 2022-07-11
fix: quarterback auth, restore comments partially, update transform, editor, quarterback

teemukoiv:


atyponci:
Can one of the admins verify this patch?

PR #79, opened 2022-08-21
UI changes for new Track Changes - LEAN-960

IslamJMomani:
- Add inline accept/reject buttons.

The designs: https://www.figma.com/file/vspHhzKTbwSkZWQWOcmIw2/Lean-workflow---Final-UI-views?node-id=11887%3A431667

### DIFF
--- a/src/components/projects/lean-workflow/EditorElement.tsx
+++ b/src/components/projects/lean-workflow/EditorElement.tsx
@@ -169,9 +169,7 @@ const EditorElement: React.FC<Props> = ({ editor }) => {
       const action = button.getAttribute('data-action')
       const changeId = button.getAttribute('data-changeid')
 
-      if (!action || !changeId) {
-        return
-      } else {
+      if (action && changeId) {
         const change = changeSet.changes.find((c) => c.id == changeId)
         if (change) {
           if (action === 'accept') {
--- a/src/components/projects/lean-workflow/EditorElement.tsx
+++ b/src/components/projects/lean-workflow/EditorElement.tsx
@@ -39,11 +39,11 @@ import { Node as ProsemirrorNode } from 'prosemirror-model'
 import { NodeSelection, Transaction } from 'prosemirror-state'
 import React, { useCallback, useState } from 'react'
 import { useDrop } from 'react-dnd'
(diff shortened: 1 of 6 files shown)


IslamJMomani:
> So can I as a local user accept my own changes anymore?

In order to do that you should have permission to `handleSuggestion` as in
https://github.com/Atypon-OpenSource/manuscripts-style-guide/blob/543b17d63b41b7e875680d78cd119de2f9f1597c/src/lib/capabilities.tsx#L99



> Had to comment out FilesDropdown component inside FigureElement since it kept crashing

Is this still happening after moving the UI to manuscripts-editor?

TeemuKoivisto:
> In order to do that you should have permission to `handleSuggestion` as in https://github.com/Atypon-OpenSource/manuscripts-style-guide/blob/543b17d63b41b7e875680d78cd119de2f9f1597c/src/lib/capabilities.tsx#L99

Ok, i see I can still adjust them locally so np.

> Is this still happening after moving the UI to manuscripts-editor?

I think it was because I had a file attachment that didnt exist. I'm atm having problems running lw-dashboard locally so cant verify but I'm pretty sure it's unrelated.

So once the editor changes are deployed, I think only things missing are removing `import '@manuscripts/track-changes-plugin/src/styles.css'` from TrackChangesPanel (if it's even used anymore). And then adding the import for `track-styles.css` somewhere, wherever they are imported.

TeemuKoivisto (approved):

src/components/projects/lean-workflow/EditorElement.tsx
```diff
@@ -124,6 +131,60 @@ const EditorElement: React.FC<Props> = ({ editor }) => {
...
+        return
+      }
+      const { changeSet } = trackState
+
+      const action = button.getAttribute('data-action')
+      const changeId = button.getAttribute('data-changeid')
+
+      if (!action || !changeId) {
+        return
+      } else {
```
TeemuKoivisto, line 174: This else is unneeded, the previous return already stops execution. Although it might work out better to negate the expression and put the else block inside the if

teemukoiv (approved):

PR #93, opened 2022-09-06
LW-R7: Production notes aren't added Correctly (LEAN-1573)

asouqi:


src/postgres-data/buildData.ts
```diff
@@ -348,6 +357,7 @@ export default async function buildData(
     user,
     api
   )
+  const productionNotes = await getProductionNotes(projectID, manuscriptID, api)
```
mbartenev-atypon, line 360: Do we really need to request them separately? Are they not coming with all the other modes and then separated into its own array? See getManuscriptData function for reference.

mbartenev-atypon (changes requested):

src/postgres-data/buildData.ts
asouqi: `/load` endpoint will not return by default notes or comments, @hhusban can we make the `/load` endpoint return notes and comments with other models?

mbartenev-atypon (approved):

mbartenev-atypon (approved):

PR #95, opened 2022-09-08
fix: re-authenticate to quarterback incase jwt is missing, refactor old track components

TeemuKoivisto:


src/quarterback/api/methods.ts
```diff
@@ -29,8 +29,16 @@ const DEFAULT_HEADERS = {
   'Content-Type': 'application/json',
 }
 
+let debouncedAuth: ReturnType<typeof setTimeout> | undefined
+
 function getAuthHeader() {
   const jwt = useAuthStore.getState().jwt
+  if (!jwt && !debouncedAuth) {
+    useAuthStore.getState().authenticate()
```
IslamJMomani, line 37: Don't you want to re assign (get) the JWT value after this step? or maybe it's fine as the second attempt to getAuthHeader should work.

IslamJMomani (approved):

src/quarterback/api/methods.ts
TeemuKoivisto: Yeah the request will fail but I didnt want to complicate it too much. After all it should work in the first place 😅

PR #99, opened 2022-09-12
Create a snapshot upon the task completion - LEAN-1587

IslamJMomani:
This is needed to resolve `Approved/Rejected suggestion displays for all users`.
So, we are moving the logic of creating a snapshot to the frontend side.


src/components/projects/lean-workflow/ManualFlowTransitioning.tsx
```diff
@@ -20,9 +20,11 @@ import {
   TaskStepDoneIcon,
   usePermissions,
 } from '@manuscripts/style-guide'
+import { trackCommands } from '@manuscripts/track-changes-plugin'
```
mbartenev-atypon, line 23: hm, I thought we are trying to move away from using that component. Am I confusing it with something?

mbartenev-atypon (approved):

src/components/projects/lean-workflow/ManualFlowTransitioning.tsx
IslamJMomani: I can't remember any discussion around that 🤔

PR #118, opened 2022-10-19
LW - R5: Block comment icon displays as inline icon (LEAN-1016)

asouqi:


src/components/projects/lean-workflow/CommentList.tsx
```diff
@@ -121,11 +131,68 @@ export const CommentList: React.FC<Props> = ({ selected, editor }) => {
...
         // newComment.originalText = getHighlightText(highlight, state)
         newComment.originalText = highlight.text
-        setNewComment(newComment)
       }
+      setNewComment(newComment)
     }
   }, [commentTarget, doc, newComment, state, currentUser])
 
+  const commentsLabel = useMemo(() => {
+    let labels: { [p: string]: string } = {}
```
mbartenev-atypon, line 140: Perhaps we can use Map here? I see it is unpacked on line 145 in the descendants iteration but it is not exactly clear to me what is the value of avoiding mutating it. It is of course not critical.

src/components/projects/lean-workflow/CommentList.tsx
```diff
@@ -121,11 +131,68 @@ export const CommentList: React.FC<Props> = ({ selected, editor }) => {
       if (highlight) {
         // newComment.originalText = getHighlightText(highlight, state)
         newComment.originalText = highlight.text
-        setNewComment(newComment)
       }
+      setNewComment(newComment)
     }
   }, [commentTarget, doc, newComment, state, currentUser])
 
+  const commentsLabel = useMemo(() => {
```
mbartenev-atypon, line 139: please add some comments to this function explaining what is going on inside so it will be easier to work with it in future, thanks

mbartenev-atypon: Also, if I understand correctly, this function collects all the labels from the doc so perhaps it should be called commentsLabels or getCommentsLabels?

mbartenev-atypon (approved):

src/components/projects/lean-workflow/CommentList.tsx
asouqi: oh that makes sense to use a map

src/components/projects/lean-workflow/CommentList.tsx
asouqi: added a comment to clarify it more, and update the name to `commentsLabels`

PR #124, opened 2022-11-11
LEAN-1695 - LW-R7: An error message appeared when opening the editor …

jirikaiser-atypon:
…for some submissions


src/components/track-changes/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+import { ManuscriptNode } from '@manuscripts/manuscript-transform'
+import {
+  CHANGE_OPERATION,
+  CHANGE_STATUS,
+} from '@manuscripts/track-changes-plugin'
+
+const hasPendingOrRejectedChanges = (node: ManuscriptNode) => {
+  if (node?.attrs?.dataTracked) {
+    const trackedEntries = node.attrs.dataTracked.filter(
+      (data: any) =>
```
mbartenev-atypon, line 21: Woudl that be possible to reduce amount of "any" we use in that function?

jirikaiser-atypon: > Would that be possible to reduce amount of "any" we use in that function?

Updated.

mbartenev-atypon (approved):

PR #129, opened 2022-11-25
LEAN-1918 Inspector Model Management in Tracked Data

mbartenev-atypon:


mbartenev-atypon:
@asouqi, please note that I removed the merging of the final modelMap and the one that is derived from the track-changes because it may reintroduce the models that are already deleted. We might need to merge only those models that are non-PM.

src/hooks/use-tracked-model-management.ts
```diff
@@ -0,0 +1,111 @@
...
+  doc: ManuscriptNode,
+  view: ManuscriptEditorView | undefined,
+  state: EditorState<ManuscriptSchema>,
+  dispatch: (tr: Transaction<any>) => EditorState<ManuscriptSchema>,
+  saveModel: <T extends Model>(model: T | Build<T> | Partial<T>) => Promise<T>,
+  deleteModel: (id: string) => Promise<string>
+) => {
+  const modelMap = useMemo(() => {
+    const docClean = filterNodesWithTrackingData(doc.toJSON())
+    return encode(schema.nodeFromJSON(docClean))
```
asouqi, line 40: we need `MPSupplement` objects for the supplement tap, I don't know if we could include them here.

mbartenev-atypon: Added supplements. I think it's ok to add any models that are not representable by PM.

asouqi (approved):

PR #134, opened 2022-12-08
LW - R7: Already cited items have the wrong icon and can be selected more than once (LEAN-1922)

asouqi:
In this PR, we make sure that we don't show the same reference item.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@manuscripts/article-editor",
-  "version": "0.2.37",
+  "version": "0.2.38-LEAN-1922",
   "license": "CPAL-1.0",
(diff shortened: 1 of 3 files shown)


mnatsheh (changes requested):
I think that this should be part of the `filterLibrary` callback we already pass; no need for an extra argument.

asouqi:
>I think that this should be part of the filterLibrary callback we already pass; no need for an extra argument.

in this case, we need to pass the citation to this callback to know which reference has been used in the citation. and this requires changing the method parameter list to include citations in the `manuscript-editor`

mnatsheh (approved):

PR #136, opened 2022-12-15
LW - R7: Add new citation on selected text takes some time to load the cited items

asouqi:
Ui fix requested in [figma](https://www.figma.com/file/vspHhzKTbwSkZWQWOcmIw2/LW---Editor-%26-Roles?node-id=16954%3A450289&t=HYp2RZefkdw0fWqH-1), and fix debounce for the search bar


src/components/library/CitationSearch.tsx
```diff
@@ -206,9 +224,13 @@ export const CitationSearch: React.FC<{
...
   }, [handleCite, selected])
 
-  const handleQuery = useCallback((event: ChangeEvent<HTMLInputElement>) => {
-    setQuery(event.target.value)
-  }, [])
+  const debouncedCallback = debounce((value) => setQuery(value.trim()), 800)
+
+  const handleQuery = useCallback(
+    (event: ChangeEvent<HTMLInputElement>) =>
+      debouncedCallback(event.target.value),
```
mbartenev-atypon, line 231: That's a rather generic and not very meaningful naming

src/components/library/CitationSearch.tsx
```diff
@@ -126,6 +128,26 @@ export const CitationSearch: React.FC<{
...
+    setCancelTokenSource,
+  ] = useState<CancelTokenSource>()
+
+  const apiSearchCallback = useCallback(
+    (
+      query: string,
+      params: { rows: number; sort?: string },
+      mailto: string
+    ) => {
+      cancelTokenSource?.cancel()
```
mbartenev-atypon, line 142: What really happens here? Do previous requests slow down the server response somehow?

mbartenev-atypon (approved):

src/components/library/CitationSearch.tsx
mbartenev-atypon: That can be fixed while in qa. Resolving

src/components/library/CitationSearch.tsx
mbartenev-atypon: Resolving to unblock merging.

asouqi: > What really happens here?

We need to cancel old requests, so when the user clicks to show more results it will be the only request to get references. and if we leave the old requests and they finish after showing more requests it's going to ruin the references list

src/components/library/CitationSearch.tsx
asouqi: do you have naming in your mind, so we could use it

mbartenev-atypon: how about debouncedSetQuery?

asouqi: renamed

PR #147, opened 2023-01-29
fix track changes hook call

asouqi:


mbartenev-atypon (approved):

mbartenev-atypon:
@asouqi , is there any specific error that is fixed by that? Something visible in the console or a specific behaviour?

asouqi:
> is there any specific error that is fixed by that? Something visible in the console or a specific behaviour?

this is related to the issue in this [PR](https://github.com/atypon/leanworkflow-dashboard/pull/430#issuecomment-1402860726) when enabling track changes, look at the [view](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/components/track-changes/useEditorStore.ts#L36) in useEditorStore

mbartenev-atypon:
> > is there any specific error that is fixed by that? Something visible in the console or a specific behaviour?
> 
> this is related to the issue in this [PR](https://github.com/atypon/leanworkflow-dashboard/pull/430#issuecomment-1402860726) when enabling track changes, look at the [view](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/components/track-changes/useEditorStore.ts#L36) in useEditorStore

Ah, I see. I added a different fix for it here: https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/various-fixes-test/src/components/projects/lean-workflow/ManuscriptPageContainerLW.tsx#L173. It look to me that the order of execution has changes with the new react version so I moved the init action to be earlier to avoid any related issues. What do you think?

asouqi:
> It look to me that the order of execution has changes with the new react version so I moved the init action to be earlier to avoid any related issues. What do you think?

that will be better.

PR #166, opened 2023-03-16
Fixing the issues "e.cancelToken.subscribe is not a function" on citation Pop up (LEAN-2281)

asouqi:
Update the version of `axios` to be compatible with body-editor


mnatsheh:
The pipeline is failing for this.

asouqi:
> The pipeline is failing for this.

fixed

PR #172, opened 2023-03-27
LEAN-2309: AMI-030 - Re-enable Authors - Implementation

nghazzawi-atypon:


asouqi (changes requested):

src/postgres-data/buildUtilities.ts
```diff
@@ -132,6 +132,26 @@ const buildUtilities = (
...
+        const authorsAndAffiliations = data.authorsAndAffiliations
+        const authors = data.authorsAndAffiliations.authors.map(
+          (author: { _id: string }) =>
+            author._id == model._id ? model : author
+        )
+
+        authorsAndAffiliations.authors = authors
+        if (authorsAndAffiliations.affiliations.get(model._id)) {
+          authorsAndAffiliations.affiliations.set(model._id, model)
+        }
```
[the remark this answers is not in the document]
nghazzawi-atypon: I faced an issue while using store inside a class component, therefore I relocated it to AuthorsModals.ts file.

asouqi (approved):

nghazzawi-atypon:
@asouqi  Could you kindly review my latest changes ? I have made updates to fix other issues with removing authors and updating affiliations. Thank you.

mnatsheh:
@nghazzawi-atypon @asouqi , why are we having to update the `authorsAndAffiliations` values ourselves? Shouldn't the functions we pass already take care of this?

nghazzawi-atypon:
@mnatsheh 
the authorsAndAffiliations has a separate state in the store that needs to be updated, while the functions we pass call the generic  function "saveModel" from buildUtilites which only update the modelMap state.

As stated in the ticket, there is an issue with the functionality of adding affiliations and authors. Therefore, we will need to create a separate ticket to fix this issue before we can proceed with resolving the current one. Please refer to the attached record: 
https://user-images.githubusercontent.com/93189633/228983327-c8860db7-1a3f-4cac-ae8f-7ae1d77182c9.mp4

asouqi (changes requested):

mbartenev-atypon:
@nghazzawi-atypon , let's make sure that the changes on the modelMap will automatically trigger an update on the data derived from it. I create a draft for that to accommodate that need in the store logic. https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/178/files. Using this it should be enough just to update the modelMap and not update the store. I think saveModel function actually updates modelMap so using saveModel should be enough.

src/components/metadata/AuthorsModals.tsx
```diff
@@ -93,49 +93,125 @@ export const AuthorsModal: React.FunctionComponent<AuthorsProps> = ({
...
+}) => {
+  const [{ authorsAndAffiliations, storedContributorRoles }, dispatch] =
+    useStore((store) => {
+      return {
+        authorsAndAffiliations: store.authorsAndAffiliations,
+        storedContributorRoles: store.contributorRoles,
+      }
+    })
+  // handle save author state
+  const handleSaveAuthorSate: (values: AuthorValues) => Promise<void> = async (
```
mbartenev-atypon, line 105: please fix typo to be "State" instead of "Sate"

nghazzawi-atypon:
@mbartenev-atypon 
I have tested your code and found an issue with the "prev state". The problem is that it is identical to the new state, causing the condition to always evaluate to false.

I have investigated this issue, but could not find any reason for it. Do you have any ideas?

nghazzawi-atypon:
@mbartenev-atypon 
The state is continuously being updated in some way that I am unable to determine. I have tested this both with and without your changes, prior to passing the state to be updated(already being updated), so I'm not able to catch the previous state ,please check the attached records (without/with your changes)

I believe that Your logic is  make more sense since we are updating the state automatically without manually manipulating the authorsAndAffliations state. However, I have a suggestion that could help us here by moving the logic I have created inside AuthorsModal.tsx to AuthorsModalContainer.tsx. We can achieve this by first converting AuthorsModalContainer.tsx into a functional component, which already contains the necessary methods like handleSaveAuthor, removeAuthor, contributorRole, and others. This will eliminate the need to redefine (replicate) these methods inside the current file. Please let me know what do you think.

https://user-images.githubusercontent.com/93189633/231261633-f9b8502b-0ad0-400c-b107-8e9f860ad50b.mp4


https://user-images.githubusercontent.com/93189633/231261661-4ef8a394-eaf7-4d56-a271-e22911cc054a.mp4

src/postgres-data/PsSource.ts
```diff
@@ -50,7 +51,18 @@ export default class PsSource implements StoreDataSourceStrategy {
     }
     next({ ...state, ...this.data, ...this.utilities })
   }
-  afterAction: StoreDataSourceStrategy['afterAction'] = (state, setState) => {
+  afterAction: StoreDataSourceStrategy['afterAction'] = (
+    state,
+    prev,
+    setState
+  ) => {
+    //  please check if that's equality check is good enough
```
mbartenev-atypon, line 59: please remove that comment, it was needed during the discussion

mbartenev-atypon (changes requested):

src/postgres-data/PsSource.ts
nghazzawi-atypon: done

mbartenev-atypon (approved):

asouqi (approved):

nghazzawi-atypon:
@mbartenev-atypon  could you please re-review my latest changes in order to fix the below issues:
https://jira.atypon.com/browse/LEAN-2385
https://jira.atypon.com/browse/LEAN-2448
https://jira.atypon.com/browse/LEAN-2449
https://jira.atypon.com/browse/LEAN-2309

PR #182, opened 2023-04-11
LEAN-2342

asouqi:


mbartenev-atypon (approved):

mbartenev-atypon (approved):

src/hooks/use-create-editor.tsx
```diff
@@ -70,6 +71,10 @@ export const useCreateEditor = () => {
   }))
   const { user: trackUser } = useAuthStore()
 
+  const getCapabilities = memoize((project, user, permittedActions) =>
```
mbartenev-atypon, line 74: just wondering, why use that and not useMemo?

asouqi: I tried it at first but didn't work it keeps recalculating them, I think it's related to prosemirror not working with React as we use that function in the props we pass to it!

PR #186, opened 2023-04-19
LEAN-2306: AMI-020 - Remove invalid attribute - Implementation

nghazzawi-atypon:


asouqi (changes requested):
please check `yarn.lock`

PR #196, opened 2023-05-02
LEAN-2309: AMI-030 - Re-enable Authors - Implementation

nghazzawi-atypon:


mbartenev-atypon (approved):

nghazzawi-atypon:
@mbartenev-atypon 
Based on our conversation, I came across an issue related to the affiliation section. Specifically, when you modify any of the author values, the selected affiliation disappears. Upon investigation, I discovered that the "AuthorValues" interface in the styleguide/types.ts file is missing the "affiliations" list. This is causing the selected values to be removed whenever an author is updated.
I'm uncertain whether I should include this value in the interface because when I attempted to do so, I encountered some errors. Please advise me on what you think I should do.
image

PR #206, opened 2023-05-22
LEAN-2438: LW-R9: Clicking Comment marker doesn't open the inspector …

mvidtman:
…and show the relevant comment

### DIFF
--- a/src/components/Panel.tsx
+++ b/src/components/Panel.tsx
@@ -20,6 +20,7 @@ import React, { useEffect, useLayoutEffect, useRef, useState } from 'react'
 import styled from 'styled-components'
 
 import layout, { Pane } from '../lib/layout'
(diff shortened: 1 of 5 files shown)


src/components/Panel.tsx
```diff
@@ -105,17 +113,18 @@ const Panel: React.FC<PanelProps> = (props) => {
...
     setState((state) => ({
       ...state,
       originalSize: size,
       size: collapsed ? 0 : size,
       collapsed,
     }))
+
+    if (forceOpen === false && data.collapsed === true) {
+      dispatch({ selectedComment: undefined })
+    }
```
asouqi, line 127: it should work without this change, can you please remove it and check if it's working

mvidtman: It does not work without this...

PR #225, opened 2023-06-25
LEAN-2320: Ceres-Stag - published XML does not include changes

mbetamony:
Based on what I read, I believe that `setTimeout` is what's causing the issue of submitting before saving
the `submit` call will not wait for the code inside the `setTimeout` callback to execute. The `setTimeout` function introduces an asynchronous delay, and the subsequent code, including the submit call, will continue executing without waiting for the delay to complete. 
I'm not why its there, but if we need to keep it we need to at least do this
`return async () => {
  const resp = await saveSnapshot(docToJSON())
  if ('data' in resp) {
    execCmd(trackCommands.applyAndRemoveChanges())
    await new Promise((resolve) => {
      setTimeout(() => {
        const state = useEditorStore.getState().editorState
        if (!state) {
          return
        }
        resolve(usePouchStore.getState().saveDoc(getDocWithoutTrackContent(state)))
      })
    })
  }
}`

### DIFF
--- a/src/hooks/use-handle-snapshot.ts
+++ b/src/hooks/use-handle-snapshot.ts
@@ -28,13 +28,11 @@ export const useHandleSnapshot = (storeExists = true) => {
     const resp = await saveSnapshot(docToJSON())
     if ('data' in resp) {
       execCmd(trackCommands.applyAndRemoveChanges())
-      setTimeout(() => {
-        const state = useEditorStore.getState().editorState
-        if (!state) {
-          return
-        }
-        usePouchStore.getState().saveDoc(getDocWithoutTrackContent(state))
-      })
+      const state = useEditorStore.getState().editorState
+      if (!state) {
+        return
+      }
+      await usePouchStore.getState().saveDoc(getDocWithoutTrackContent(state))
     }
   }
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@manuscripts/article-editor",
-  "version": "1.1.20",
+  "version": "1.1.20-LEAN-2320B",
   "license": "CPAL-1.0",
   "description": "React components for editing and viewing manuscripts",
   "repository": "github:Atypon-OpenSource/manuscripts-article-editor",


mbartenev-atypon:
Indeed, it seems, the submitting of task will not wait for the frontend to save the data and the 'handleSnapshot' will race 'submit', which can explain why it some times work ok and some times doesn't. I looked a bit in why the setTimeout is there and it looks like the problem is in the setActions being event based and essentially also asynchronous, but without any way to check for completion. So setTimeout is there as a kludge for that problem - to make sure it will be stacked and called after the actions executions.
The proposed solution looks good to me.

mbartenev-atypon (approved):

src/hooks/use-handle-snapshot.ts
```diff
@@ -28,13 +28,11 @@ export const useHandleSnapshot = (storeExists = true) => {
     const resp = await saveSnapshot(docToJSON())
     if ('data' in resp) {
       execCmd(trackCommands.applyAndRemoveChanges())
-      setTimeout(() => {
```
mbartenev-atypon, line 31: Can you add the code you supplied in the comments to be used here?

mbartenev-atypon (changes requested):

src/hooks/use-handle-snapshot.ts
mbetamony: Sure, done.

mbetamony:
rerun

src/hooks/use-handle-snapshot.ts
```diff
@@ -28,12 +28,16 @@ export const useHandleSnapshot = (storeExists = true) => {
...
-        const state = useEditorStore.getState().editorState
-        if (!state) {
-          return
-        }
-        usePouchStore.getState().saveDoc(getDocWithoutTrackContent(state))
+      await new Promise((resolve) => {
+        setTimeout(() => {
+          const state = useEditorStore.getState().editorState
+          if (!state) {
+            return
```
mbartenev-atypon, line 35: all good by maybe here we can resolve or reject too, since we are in a promise now

mbartenev-atypon (approved):

src/hooks/use-handle-snapshot.ts
mbetamony: should it throw an error? 
 reject(new Error('State is not available'))

mbartenev-atypon: yes, looks good

PR #228, opened 2023-07-04
LEAN-2505 (Implementation of multiple pending comments)

asouqi:
In this PR will do the following:
* replace the pending comment in the store to be a map of comments with their ids
* move CRUD callback for comment to a hook, so we could test them later in [PR#177](https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/177) to avoid regression.


src/hooks/use-comment-manger.ts
```diff
@@ -0,0 +1,201 @@
+/*!
```
mbartenev-atypon, line 1: there is a type in the file name

mbartenev-atypon: I don't see any usage of this file and it doesn't seem to be exported out of index to be used anywhere else. Do we still needed?

src/hooks/use-create-editor.tsx
```diff
@@ -117,7 +117,14 @@ export const useCreateEditor = () => {
     modelMap,
     getManuscript: () => manuscript,
     getCurrentUser: () => user,
-    setComment: (comment?: CommentAnnotation) => dispatch({ comment }),
+    setComment: (comment?: CommentAnnotation) => {
+      if (comment) {
+        const state = getState()
+        const contribution = buildContribution(state.user._id)
+        comment.contributions = [contribution]
+        dispatch({ newComments: state.newComments.set(comment._id, comment) })
```
mbartenev-atypon, line 125: Why do we need to mutate existing comments map? Can we create a new one so if there is something depending on an update of it, it will be able to react?

src/components/projects/Inspector.tsx
```diff
@@ -58,7 +58,7 @@ const Inspector: React.FC<Props> = ({ editor }) => {
     trackModelMap: store.trackModelMap,
     attachments: store.attachments,
     fileManagement: store.fileManagement,
-    editorSelectedComment: store.comment,
+    editorSelectedComment: store.newComments.size,
```
mbartenev-atypon, line 61: This is a bit confusing now. We have a "editorSelectedComment" property, which value now is integer. Would that be possible to rename appropriately?

src/hooks/use-comment-manger.ts
```diff
@@ -0,0 +1,201 @@
...
+    const commentsTreeMap = buildCommentTree(doc, combinedComments)
+    return Array.from(commentsTreeMap.entries())
+  }, [trackModelMap, newCommentsList, doc])
+
+  const removePendingComment = useCallback(
+    (comment: CommentAnnotation) => {
+      if (newComments.has(comment._id)) {
+        newComments.delete(comment._id)
+        dispatch({
+          newComments,
```
mbartenev-atypon, line 79: Here is mutation of a property in the store again. This won't trigger an update for components that take 'newComments' from the store. A new map should be created. E.g.: new Map(newComments)

mbartenev-atypon (changes requested):

src/hooks/use-comment-manger.ts
asouqi: updated

src/components/projects/Inspector.tsx
asouqi: updated

src/hooks/use-create-editor.tsx
asouqi: updated

src/hooks/use-comment-manger.ts
asouqi: we use it in https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/5804e9252fb2d1451419ea26fd563699772ca61d/src/components/projects/CommentList.tsx#L34

mbartenev-atypon: ah, It was collapsed for me. Can you please fix the typo? "-manger" to "...-manager"

src/components/projects/HighlightedText.tsx
```diff
@@ -25,11 +30,25 @@ const Container = styled.div`
...
+  commentsLabels: Map<string, string>
   onClick?: (comment: CommentAnnotation) => void
-}> = React.memo(({ comment, getHighlightTextColor, onClick }) => {
-  if (!comment.originalText) {
+}> = React.memo(({ comment, state, commentsLabels, onClick }) => {
+  const getHighlightTextColor = useCallback(
+    (comment: CommentAnnotation) =>
+      !isHighlightComment(comment) ||
+      (state && getHighlightTarget(comment, state))
+        ? '#ffe08b'
```
mbartenev-atypon, line 42: It would be ideal to grab those colors from theme if they are there.

src/hooks/use-comment-manger.ts
asouqi: updated

mbartenev-atypon (approved):

PR #230, opened 2023-07-11
LEAN-2035 (update react-select)

asouqi:


asouqi:
@mnatsheh do we have a green light to update `react-select` for other repo(`manuscripts-body-editor`, `manuscripts-comment-editor`, `manuscripts-comment-editor`) or that will be done in another ticket?

mnatsheh:
@asouqi yes, please go ahead.

src/components/projects/CategorisedKeywordsInput.tsx
```diff
@@ -316,7 +316,7 @@ export const CategorisedKeywordsInput: React.FC<{
       <OptionWrapper
         focused={keywordToEdit ? data._id === keywordToEdit._id : false}
         ref={innerRef}
-        {...innerProps}
+        {...(innerProps as any)}
```
mbartenev-atypon, line 319: any chance we can do a more specific type here?

mbartenev-atypon (approved):

src/components/projects/CategorisedKeywordsInput.tsx
asouqi: updated the ref to be null to ignore the ref of innerProps so with that no need for the `any`. 

and I used it as there is a type mismatch with the ref of styled-components dependency and the one we get from `react-select`. in the new version of [react-select](https://github.com/JedWatson/react-select/blob/master/docs/styled-components.tsx) they remove that dependency and use inline CSS to style custom component!

PR #231, opened 2023-07-11
LEAN-2615: update dependencies

mbetamony:


mbetamony:
I will update the body-editor once it finishes from Jenkins

PR #237, opened 2023-07-19
LEAN-2320 Fix for racing on saving manuscripts backend and proceed mutation

mbartenev-atypon:


mbartenev-atypon:
@mbetamony please check. This is partially based on your work

mbetamony (approved):

src/hooks/use-handle-snapshot.ts
```diff
@@ -28,12 +28,23 @@ export const useHandleSnapshot = (storeExists = true) => {
     const resp = await saveSnapshot(docToJSON())
     if ('data' in resp) {
       execCmd(trackCommands.applyAndRemoveChanges())
-      setTimeout(() => {
-        const state = useEditorStore.getState().editorState
-        if (!state) {
-          return
-        }
-        usePouchStore.getState().saveDoc(getDocWithoutTrackContent(state))
+      return new Promise<void>((resolve, reject) => {
```
mbetamony, line 31: @mbartenev-atypon I'm not sure if this why Nick is having save issues even when he's in the PE step. 
but it might be related to this
I'm not sure if this actually fixes the race 
a couple of solutions code be 
``` 
return async () => {
    const resp = await saveSnapshot(docToJSON())
    if ('data' in resp) {
      execCmd(trackCommands.applyAndRemoveChanges())
      return new Promise<void>((resolve, reject) => {
        setTimeout(() => {
          const state = useEditorStore.getState().editorState
          if (!state) {
            reject(new Error('State is not available'))
            return
          }
          usePouchStore
            .getState()
            .saveDoc(getDocWithoutTrackContent(state))
            .then(() => {
              resolve()
            })
            .catch(() =>
              reject(new Error('Cannot save to api. Check connection.'))
            )
        }, 1000) // to avoid potentially saving before the changes are applied
      })
    }
  }
  ```
  OR 
  ```
  return async () => {
  const resp = await saveSnapshot(docToJSON())
  if ('data' in resp) {
    execCmd(trackCommands.applyAndRemoveChanges())
    return new Promise<void>((resolve, reject) => {
      setTimeout(() => {
        const state = useEditorStore.getState().editorState
        if (!state) {
          reject(new Error('State is not available'))
          return
        }
        return usePouchStore
          .getState()
          .saveDoc(getDocWithoutTrackContent(state))
          .then(() => {
            resolve()
          })
          .catch(() =>
            reject(new Error('Cannot save to api. Check connection.'))
          )
      }, 1000) // to avoid potentially saving before the changes are applied
    })
  }
}
  ```
  I also tried to upload a package on dev3 and everything worked as expected, but this might be worth to take a look at

PR #272, opened 2023-09-18
LEAN-2859

mnatsheh:


mnatsheh:
Based on discussion with Mikhail, I switched back to using the original model map. Although it does solve the sync problem, it's not definite that the model map is complete, and we should be moving away from using model map in body-editor anyway. I'm assuming the same applies to Ahmad's change.

asouqi:
> Based on discussion with Mikhail, I switched back to using the original model map. Although it does solve the sync problem, it's not definite that the model map is complete, and we should be moving away from using model map in body-editor anyway. I'm assuming the same applies to Ahmad's change.

I will revert that from my side. @mnatsheh for showing the reference list model will still build a model map that is derived from PM document is that ok?

mnatsheh:
> > Based on discussion with Mikhail, I switched back to using the original model map. Although it does solve the sync problem, it's not definite that the model map is complete, and we should be moving away from using model map in body-editor anyway. I'm assuming the same applies to Ahmad's change.
> 
> I will revert that from my side. @mnatsheh for showing the reference list model will still build a model map that is derived from PM document is that ok?

As long as we don't encode the whole doc, I'm ok with that.

mbartenev-atypon (approved):

PR #276, opened 2023-09-20
REFTC-90 - Finalize TC in prosemirror (LEAN-2737)

asouqi:


src/components/track-changes/utils.ts
```diff
@@ -100,7 +106,8 @@ export const adaptTrackedData = (docJSONed: unknown) => {
           // @ts-ignore
           child.attrs = deepCloneAttrs(child.attrs) || {} // @TODO: needs refactoring, in case when there is a dataTracked attribute, we deep copy attributes 2 times.
           // @ts-ignore
-          child.attrs.id = child.attrs.id + trackedJoint + lastChange.id
+          child.attrs[child.attrs.id ? 'id' : 'rid'] =
```
mbartenev-atypon, line 109: Are we sure that when there is no id, there is always rid?

asouqi: I double check all the [nodes](https://github.com/Atypon-OpenSource/manuscripts-transform/tree/master/src/schema/nodes) in manuscripts-transform and will have just(id or rid)

mbartenev-atypon (approved):

PR #288, opened 2023-10-01
REFTC-100 - See the Reference modifications inline (LEAN-2723)

asouqi:


src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -170,6 +171,16 @@ const ManuscriptPageView: React.FC = () => {
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [state])
 
+  const trackAttrsPopper = useTrackAttrsPopper()
+
+  const onAppClick = useCallback(
```
mbartenev-atypon, line 176: this looks redundant,  why not just pass "trackAttrsPopper" to "onClick"?

mbartenev-atypon (approved):

src/components/projects/ManuscriptPageContainer.tsx
asouqi: I separated `trackAttrsPopper` so we could later add another callback to handle the case of figure and file inspector dropdown list

PR #316, opened 2023-12-13
Migrate Manuscript Api to V2

asouqi:
This PR will replace the calls for manuscript API to V2. and use API instead of Quarterback API


mnatsheh (commented):
Is it easy to get rid of quarterback-types as part of this? we can copy the types that we need.

asouqi:
> Is it easy to get rid of quarterback-types as part of this? we can copy the types that we need.

yes it's easy, I removed it I don't if this ok @mbartenev-atypon

mnatsheh (changes requested):
As far as I can tell, several functions under `Api` are not used so let's get rid of them:

`getProject`
`getProjectCollaborators`
`getProjectModels`
`deleteProject`
`getManuscriptModels`
`saveProject`

We can also remove the `types` param from GET requests.

There are also a bunch of apis that should not be used from FE (e.g. `createNewManuscript`), but that's a cleanup for another time.

asouqi:
> As far as I can tell, several functions under `Api` are not used so let's get rid of them:
> 
> `getProject` `getProjectCollaborators` `getProjectModels` `deleteProject` `getManuscriptModels` `saveProject`
> 
> We can also remove the `types` param from GET requests.
> 
> There are also a bunch of apis that should not be used from FE (e.g. `createNewManuscript`), but that's a cleanup for another time.

we need `saveProject` as we use it when we complete task

src/postgres-data/Api.ts
```diff
@@ -44,14 +41,13 @@ export default class Api {
     })
   }
 
-  get = async <T>(url: string, baseURL?: string) => {
+  get = async <T>(url: string, data?: unknown) => {
```
mnatsheh, line 44: Let's remove the data parameter.

mnatsheh (approved):
Looks OK to me. @mbartenev-atypon can you take a quick look here?

src/components/inspector/SnapshotsDropdown.tsx
```diff
@@ -33,10 +35,10 @@ import { useEditorStore } from '../track-changes/useEditorStore'
 
 export const SnapshotsDropdown: React.FC = () => {
   const { wrapperRef, toggleOpen, isOpen } = useDropdown()
-  const snapshotStore = useSnapshotStore()
+  const [api] = useStore((store) => store.api)
```
mbartenev-atypon, line 38: Exposing API to be something available through the store or to be available to components in general, looks to me like we overexpose it. It would be preferable if api was contained on its own architectural level and was abstracted away without being needed to components or hooks on every rerender. This is why we designed the Source strategy: we abstract APIs and data assembly away from react and prepare app to work with data in a concise and structured way.
One way we can solve this is to initiate stores on higher level (in EditorApp.tsx for example) and make sure we pass api there on initiation. But it seems to me, the most optimal solution would to to get rid of the zustand stores (snapshot store, doc store, editor state store) as a part of this PR and put all the api communication in one place.
In [LEAN-2702](https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/252/files#diff-c8f08befaeaf66733590edc1fa0590f40fa68b52f596f8ee5fdc03de32980aff) we are going to introduce a new data source for QB related data and this is something what we can take advantage of to make the put all the communcation scattered in the hook into one place.
This is especially considering that we do many changes here to those stores, while ideally they shouldn't really exist anyway.

mbartenev-atypon (changes requested):

src/components/inspector/SnapshotsDropdown.tsx
mbartenev-atypon: cc @mnatsheh

asouqi: I will wait for LEAN-2702 to be merged

src/postgres-data/Api.ts
asouqi: updated

src/components/inspector/SnapshotsDropdown.tsx
asouqi: update

src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -65,7 +48,11 @@ export function TrackChangesPanel() {
         ids.push(child.id)
       })
     }
-    execCmd(trackCommands.setChangeStatuses(CHANGE_STATUS.accepted, ids))
+    editor.view &&
+      trackCommands.setChangeStatuses(CHANGE_STATUS.accepted, ids)(
```
mbartenev-atypon, line 52: as mentioned previously execCmd was a nice abstraction, it can either leave in a hook of its own or be created in a QB source.

src/hooks/use-handle-snapshot.ts
```diff
@@ -9,43 +9,139 @@
...
+const hasChanged = (a: Model, b: Model): boolean => {
+  return !!Object.keys(a).find((key: keyof Model) => {
+    if (EXCLUDED_KEYS.includes(key)) {
+      return false
+    }
+    return !isEqual(a[key], b[key])
+  })
+}
+
+const saveDoc = async (
```
mbartenev-atypon, line 60: this function and all what it used for it (hasChanged function, EXCLUDED_KEYS, type Maybe, type Ok, type Error) is too much info for this file. useHandleSnapshot should only import saveDoc. It can live in the store if it needs to be created dynamically of move into a different api related file / folder.

mbartenev-atypon (approved):

src/hooks/use-handle-snapshot.ts
```diff
@@ -9,43 +9,139 @@
...
+    return resp
   }
 
   return async () => {
-    const resp = await saveSnapshot(docToJSON(), authToken)
-    if ('data' in resp) {
-      execCmd(trackCommands.applyAndRemoveChanges())
+    const resp = await saveSnapshot(projectID, manuscriptID)
+    if (resp && view) {
+      trackCommands.applyAndRemoveChanges()(view.state, view.dispatch)
```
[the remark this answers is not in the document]
asouqi: updated

src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -77,27 +76,27 @@ const ManuscriptPageView: React.FC = () => {
...
   useWindowUnloadEffect(undefined, preventUnload)
 
   const { state, dispatch, view } = editor
 
+  const handleSnapshot = useHandleSnapshot(view)
+
+  useEffect(() => {
+    storeDispatch({ handleSnapshot })
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [view])
```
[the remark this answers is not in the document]
asouqi: updated to use `state`

src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -147,10 +138,12 @@ const ManuscriptPageView: React.FC = () => {
...
   useEffect(() => {
+    const trackState = view
+      ? trackChangesPluginKey.getState(view.state)
+      : undefined
+
     doWithThrottle(() => {
-      // @TODO remove zustand editorState store, remove doc from store and only save entire editoreState into the store
-      setEditorState(state)
-      storeDispatch({ doc: state.doc })
+      storeDispatch({ doc: state.doc, trackState })
```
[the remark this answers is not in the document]
asouqi: hmm, I will leave it as we use just doc in the state

src/components/track-changes/TrackChangesPanel.tsx
asouqi: updated to hook

src/hooks/use-handle-snapshot.ts
asouqi: moved to `savingUtilities.ts`

PR #322, opened 2023-12-20
LEAN-2915: CTC-010 - Styling of the Inspector History tab - FE implem…

mvidtman:
…entation


src/theme/theme.ts
```diff
@@ -40,6 +40,11 @@ export const theme: DefaultTheme = {
       success: colors.peppermintGreen,
       warning: colors.butteryYellow,
       selected: colors.aliceBlue,
+      newDesign: {
```
mbartenev-atypon, line 43: Please change the name to more meaningful

src/theme/theme.ts
```diff
@@ -54,6 +59,11 @@ export const theme: DefaultTheme = {
         default: colors.mercuryGrey,
         hover: colors.manuscriptsLight,
       },
+      newDesign: {
```
mbartenev-atypon, line 62: Please change the name to something more meaningful

mbartenev-atypon (changes requested):

src/theme/theme.ts
mvidtman: changed to 'tracked'

src/theme/theme.ts
mvidtman: changed to 'tracked'

src/components/track-changes/suggestion-list/Suggestion.tsx
```diff
@@ -175,22 +176,40 @@ const Wrapper = styled.li<{
...
+  ${(props) =>
+    props.isFocused
+      ? `${Actions} {
+        visibility: visible;
+      }`
+      : ''}
 
   &:hover {
-    background: ${(props) => props.theme.colors.background.fifth} !important;
+    background: ${(props) => props.theme.colors.background.tracked?.hover};
```
mbartenev-atypon, line 210: Why is "tracked" optional? What are the circumstances under which it wouldn't exist? And if so wouldn't it require some handling because it would return 'undefined' which is not a valid css? Will styled-components take care of that?

mbartenev-atypon (changes requested):

src/theme/theme.ts
mbartenev-atypon: Great, thanks

src/components/track-changes/suggestion-list/Suggestion.tsx
mvidtman: updated as not optional...

mbartenev-atypon (approved):

PR #330, opened 2024-01-10
LEAN-2992

mnatsheh:


mbartenev-atypon:
I thought the plan was to move menus into the article editor or did I miss something?

mnatsheh:
> I thought the plan was to move menus into the article editor or did I miss something?

Since the body-editor needed the types I thought I would just put it here. An alternative is to put the types in body-editor and the business logic in article-editor (which is basically what I did for the toolbar), so I'm happy to go with that route if it's preferred.

mbartenev-atypon:
> > I thought the plan was to move menus into the article editor or did I miss something?
> 
> Since the body-editor needed the types I thought I would just put it here. An alternative is to put the types in body-editor and the business logic in article-editor (which is basically what I did for the toolbar), so I'm happy to go with that route if it's preferred.

I would probably prefer to minimise the usage of prosemirror components in the styleguide and semnatically make it clear that menus shouldn't be used anywhere else so if that's not too much trouble I'd suggest to move them to the article-editor. But this is not critical.

mbartenev-atypon (approved):

src/components/projects/ManuscriptMenus.tsx
```diff
@@ -0,0 +1,94 @@
...
+    })
+    .filter(Boolean) as (MenuSpec | MenuSeparator)[]
+}
+
+export const ManuscriptMenus: React.FC<ApplicationMenusProps> = ({
+  editor,
+}) => {
+  const can = usePermissions()
+
+  const specs = processMenu(getEditorMenus(editor), (spec: MenuSpec) => {
```
mbartenev-atypon, line 72: This calculations are complex enough to be wrapped with useMemo.

src/components/projects/ManuscriptMenus.tsx
```diff
@@ -0,0 +1,94 @@
...
+} from '@manuscripts/style-guide'
+import React from 'react'
+
+import config from '../../config'
+
+interface ApplicationMenusProps {
+  editor: ReturnType<typeof useEditor>
+}
+
+const processMenu = (
```
mbartenev-atypon, line 30: This seems to be a very generic name. Perhaps something like a permissionFilter can be more descriptive? What do you think?

src/components/projects/ManuscriptMenus.tsx
```diff
@@ -0,0 +1,94 @@
...
+        if (processed.submenu) {
+          processed.submenu = processSubmenu(processed.submenu, fn)
+        }
+        return processed
+      }
+    })
+    .filter(Boolean) as MenuSpec[]
+}
+
+const processSubmenu = (
```
mbartenev-atypon, line 47: Seems like the difference between processMenu and processSubmenu are minimal and these 2 functions can be combined into one

src/components/projects/OrderedListToolbarItem.tsx
```diff
@@ -0,0 +1,251 @@
...
+  )
+}
+
+/**
+ *  Will use node depth as an indicator for the list level, and `sharedDepth` will
+ *  give us all depths for the selection so we can find the parent list
+ *  After that will look at the nodes in the parent list and just update
+ *  the list that had the same level to the target level user has selected
+ */
+export const updateListStyle = (
```
mbartenev-atypon, line 189: In my opinion all this functionality should be in the body-editor and just imported to be used in the tsx components here.

src/components/projects/OrderedListToolbarItem.tsx
```diff
@@ -0,0 +1,251 @@
...
+      return (
+        node.type === schema.nodes.ordered_list ||
+        node.type === schema.nodes.bullet_list
+      )
+    }) + 1
+  const parentListNode = $from.node(parentListDepth)
+  const parentListPos = $from.before(parentListDepth)
+
+  if (parentListDepth === targetLevel) {
+    replaceToOrderList(
```
mbartenev-atypon, line 211: replaceToOrderList is one argument away from being a pure function declared out this function scope. I know it's copied so this is optional for this PR.

mbartenev-atypon (changes requested):

mbartenev-atypon:
In general I'm ok with this approach but it seems that the menus functionality is now spread across 3 component. Some parts are in the style-guide, some in the article-editor and some are in the style-guide. I'm not sure how beneficial it is since menus and not shared between multiple components and the usage of the stories does not appear to be very useful for us anyway. But we can get back to this sometime later after this is merged.

mbartenev-atypon:
> @mbartenev-atypon, there is no prosemirror usage in the menu logic that was moved to style-guide. I thought about it, and having stories in style-guide is why I would like to have that logic there (so it can change without having the run a full app to try it).

ah, yes I see now. Thanks

src/components/projects/OrderedListToolbarItem.tsx
mnatsheh: Agreed. I've moved the logic to body-editor and simplified it as it was convoluted for what it does.

src/components/projects/ManuscriptMenus.tsx
mnatsheh: Changed.

src/components/projects/ManuscriptMenus.tsx
mnatsheh: I've refactored the code to eliminate the duplication.

src/components/projects/ManuscriptMenus.tsx
mnatsheh: Done.

mbartenev-atypon (approved):

PR #331, opened 2024-01-11
LEAN-3083 Auhors and Affiliations to Prosemirror

mbartenev-atypon:


asouqi (approved):

src/tracked-models/adapt-tracked-data.ts
```diff
@@ -0,0 +1,111 @@
+/*!
```
mbartenev-atypon, line 1: This was restored and not originally created in this PR. Originally deleted as one of the big cleanup tickets where duplicated ids problems were also tackled.

PR #342, opened 2024-01-26
Cleanup/update dependencies

mnatsheh:
I also deleted a bunch of stuff that is no longer needed.


src/EditorApp.tsx
```diff
@@ -14,10 +14,9 @@ import React, { useEffect, useState } from 'react'
 import { BrowserRouter as Router } from 'react-router-dom'
 import styled from 'styled-components'
 
-import { ModalProvider } from './components/ModalHookableProvider'
```
mbartenev-atypon, line 17: As I pointed out on a number of occasions, we need to create separate PRs for different things. It would be ok to combine them if all the changes together formed a small PR but this is not the case here. Not only this PRs hard to check but it is hard to actually investigate why those changes were produced whenever debugging is needed. For example, while git blaming, I would be checking this commit and would try to makes sense out changes there were performed here, but that would be very laborious to understand what changes are related to dependencies updates and what changes are just cleanup. I especially insist that all _cleanup_ changes have to be filed under separate PRs because we can omit those during debugging and they are faster to review. Please note that during recent big PRs that combined a lot of clean up, refactoring and features we missed some things and a number of regressions were introduced: work on [LEAN-3168](https://jira.atypon.com/browse/LEAN-3168) and on [LEAN-3261](https://jira.atypon.com/browse/LEAN-3261) and LEAN-3262 (these are the ones of which I'm aware) is related to that.

Please separate all clean up related changes into a different PR.

mbartenev-atypon (changes requested):

src/components/metadata/AuthorsPage.tsx
```diff
@@ -13,46 +13,19 @@
 import AddedIcon from '@manuscripts/assets/react/AddedIcon'
 import AuthorPlaceholder from '@manuscripts/assets/react/AuthorPlaceholder'
 import ContributorDetails from '@manuscripts/assets/react/ContributorDetailsPlaceholder'
-import InvitationPlaceholder from '@manuscripts/assets/react/InvitationPlaceholder'
 import React from 'react'
-import { FormattedMessage } from 'react-intl'
 import styled from 'styled-components'
 
 const AddAuthorsMessage = () => (
```
IslamJMomani, line 19: I believe all these messages are not in use, and we can get rid of them.

We are not using the `AddAuthorsModalContainer` and all of its related components, including `AddAuthorsModal`, `AddAuthorsSidebar`, `AddAuthorsPage`, and other related components.

Is there a reason to keep them?

src/lib/collaborators.ts
```diff
@@ -10,29 +10,11 @@
...
-
-  return [
-    ...project.owners.map(getCollaborator),
-    ...project.writers.map(getCollaborator),
-    ...project.viewers.map(getCollaborator),
-  ].filter((_) => _)
-}
+import { UserCollaborator, UserProfile } from '@manuscripts/json-schema'
 
 export const buildCollaboratorProfiles = (
```
IslamJMomani, line 15: Do we need this?

IslamJMomani (approved):

src/EditorApp.tsx
mnatsheh: I've removed cleanup changes from this PR, which includes:
- `ManuscriptPlaceholder`.
- `Store` and related items.
- `DOIInput`.

Everything else that's left should be related to a dependency change, or deletion of unused code.

src/components/metadata/AuthorsPage.tsx
mnatsheh: There will be a separate ticket to move the author modals to style-guide, and delete unused items.

src/lib/collaborators.ts
mnatsheh: It's still used in `buildData`.

mbartenev-atypon (approved):

asouqi (approved):

PR #379, opened 2024-03-11
LEAN-3358 Alerting of unsaved changes and flushing throttle to avoid loosing latest changes

mbartenev-atypon:
- Using already established alert user is being warned while throttle is being flushed
- Store logic has been review to reduce amount of rerenders. This was needed to be done as a part of this ticket so the debouncing update won't trigger unnecessary rerenders.


mnatsheh (approved):

asouqi (approved):

mbartenev-atypon:
@mnatsheh @asouqi  please note that some more changes have been done to adjust store's logic to reduce amount of rerenders. This was needed to be done as a part of this ticket so the debouncing update won't trigger unnecessary rerenders.

mbartenev-atypon:
@mnatsheh @asouqi
I additionally change throttling to debouncing so there is no saving occurs during typing. That should provide smoother typing experience.

src/components/UtilitiesEffects.tsx
```diff
@@ -0,0 +1,20 @@
+/*!
```
mbartenev-atypon, line 1: This so the hook doesn't trigger unnecessary updates on its meaningful parent

PR #378, opened 2024-03-11
LEAN-3278 Allow writers to edit without track changes

mbartenev-atypon:

PR #381, opened 2024-03-13
LEAN-3447 Allow writers to edit without track changes

mbartenev-atypon:
Same PR as https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/378 but on a branch that corresponds to the newly created ticket.


mnatsheh (approved):

src/components/projects/Inspector.tsx
```diff
@@ -110,7 +110,7 @@ const Inspector: React.FC<Props> = ({ editor }) => {
                   key="comments"
                 />
               </InspectorTabPanel>
-              {config.quarterback.enabled && (
```
mnatsheh, line 113: The check for the tab header also needs to change (L#95)

mbartenev-atypon: Yes, thanks!

PR #392, opened 2024-03-24
LTYP-110 - Support for bulleted lists (LEAN-3423)

asouqi:


mnatsheh (commented):
Why can't we continue to use `useDropdown ` here?

asouqi:
> Why can't we continue to use useDropdown here?

reverted to use dropdown hook so we could take advantage of the outside click to close the list, and the general `contextList` state will decide when to open /close the list, which will help to avoid opening both lists

mnatsheh:
Shouldn't `useDropdown` automatically close the dropdown when the other dropdown is clicked? or am I missing something?

asouqi:
> Shouldn't useDropdown automatically close the dropdown when the other dropdown is clicked? or am I missing something?

yes that is right, but when I test that locally the outside click listener will not be triggered just for the case of clicking on the second list option menu, if we click anywhere other than that place it's going to be triggered!!!

asouqi:
> yes that is right, but when I test that locally the outside click listener will not be triggered just for the case of clicking on the second list option menu, if we click anywhere other than that place it's going to be triggered!!!

find the issue of that ignore, the stop of event [propagation](https://github.com/Atypon-OpenSource/manuscripts-style-guide/blob/master/src/hooks/use-dropdown.ts#L23) will not lead to the [outside](https://github.com/Atypon-OpenSource/manuscripts-style-guide/blob/master/src/hooks/use-dropdown.ts#L39C14-L39C30) click listener

mnatsheh (approved):

PR #409, opened 2024-04-17
LEAN-1334 UIQW-010 - Open Editor with Inspector open on Comments by d…

fwaisi:
…efault


fwaisi:
> @fwaisi, I see nothing regarding force opening the inspector!

Because it's already opened by default.

IslamJMomani:
> > @fwaisi, I see nothing regarding force opening the inspector!
> 
> Because it's already opened by default.

From what I can see in [this line](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/components/projects/Inspector.tsx#L88), it currently depends on the existence of a suggestion or comment. Are these conditions met in your case?

fwaisi:
> > > @fwaisi, I see nothing regarding force opening the inspector!
> > 
> > 
> > Because it's already opened by default.
> 
> From what I can see in [this line](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/components/projects/Inspector.tsx#L88), it currently depends on the existence of a suggestion or comment. Are these conditions met in your case?

yes, it meet the condition in my case. do we need to make it open even though there are no comments and suggestions?

fwaisi:
> > > > @fwaisi, I see nothing regarding force opening the inspector!
> > > 
> > > 
> > > Because it's already opened by default.
> > 
> > 
> > From what I can see in [this line](https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/components/projects/Inspector.tsx#L88), it currently depends on the existence of a suggestion or comment. Are these conditions met in your case?
> 
> yes, it meet the condition in my case. do we need to make it open even though there are no comments and suggestions?

@IslamJMomani, I have added a comment on the ticket for Pans to advise on it.
thanks.

IslamJMomani (approved):
Approved! Just please recheck the package.json file before merging as it doesn't look right.

IslamJMomani (approved):

PR #452, opened 2024-05-18
LEAN-3623

mnatsheh:
In this PR, a new component is introduced that manages selected suggestions. It works using the `data-track-id` attribute generated by TC, so no extra decorations are needed.  The way styling works is a bit awkward, but I think it's much better than using transactions.


src/components/Panel.tsx
```diff
@@ -72,22 +72,20 @@ const Panel: React.FC<PanelProps> = (props) => {
   const firstRender = useRef(true)
   const [store, dispatch] = useStore((store) => ({
     selectedComment: store.selectedComment,
-    selectedSuggestion: store.selectedSuggestion,
-    editorSelectedSuggestion: store.editorSelectedSuggestion,
```
IslamJMomani, line 76: Another mention of 'editorSelectedSuggestion' in line 139 needs to be removed.

src/components/track-changes/suggestion-list/Suggestion.tsx
```diff
@@ -34,53 +34,27 @@ export const Suggestion: React.FC<Props> = ({
...
-  }, [isSelectedSuggestion]) // eslint-disable-line react-hooks/exhaustive-deps
+  const isSelected = suggestion.id === selectedSuggestionID
+
+  const handleClick = (e: React.MouseEvent) => {
+    e.preventDefault()
+    dispatch({
+      selectedSuggestionID: suggestion.id,
+    })
+    setModalVisible(true)
+    handleClickSuggestion(suggestion)
```
IslamJMomani, line 52: Would this just set the selected suggestion ID? But L48 already does that.

src/components/track-changes/suggestion-list/Suggestion.tsx
```diff
@@ -34,53 +34,27 @@ export const Suggestion: React.FC<Props> = ({
...
-    <Wrapper
-      data-cy="suggestion"
-      isFocused={isSelectedSuggestion}
-      ref={wrapperRef}
-    >
-      <FocusHandle
-        href="#"
-        onClick={(e) => {
-          e.preventDefault()
-          if (suggestion.dataTracked.status !== CHANGE_STATUS.rejected) {
```
IslamJMomani, line 77: Did you intentionally remove this check?

IslamJMomani (approved):

mbartenev-atypon (changes requested):

src/components/SelectionStyles.tsx
```diff
@@ -0,0 +1,60 @@
...
+import { scrollIntoViewIfNeeded } from '../lib/scroll'
+import { useStore } from '../store'
+
+export const SelectionStyles: React.FC = () => {
+  const [{ selectedSuggestionID }, dispatch] = useStore((state) => ({
+    selectedSuggestionID: state.selectedSuggestionID,
+  }))
+
+  useEffect(() => {
+    document.addEventListener('click', (e) => {
```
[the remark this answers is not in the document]
mnatsheh: If we use editor state to track this, then every node with a change will have to be re-rendered every time this state changes (i.e. on every click). This might not be an issue for most nodes, but it is for nodes like like `bibliography_element` for which the rendering is expensive. We can add logic to improve the performance for these cases, but I think that's too complicated just to achieve some style change.

mbartenev-atypon, line 23: event listener should be unsubscribed on ummount in useEffect

src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -10,179 +10,112 @@
...
+    const state = editor.state
+    const view = editor.view
+    const tr = state.tr
+    if (suggestion.type === 'text-change') {
+      const pos = suggestion.to - 1
+      tr.setSelection(TextSelection.create(state.doc, pos, pos))
+    } else {
+      tr.setSelection(NodeSelection.create(state.doc, suggestion.from))
     }
+    view?.dispatch(tr.scrollIntoView())
```
asouqi, line 47: will need to also add `view.focus()` for that to work when the selection is out of editor

src/lib/scroll.ts
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * All portions of the code written by Atypon Systems LLC are Copyright (c) 2024 Atypon Systems LLC. All Rights Reserved.
+ */
+
+export const scrollIntoViewIfNeeded = (element: Element) => {
+  const rect = element.getBoundingClientRect()
+  if (rect.bottom > window.innerHeight || rect.top < 150) {
+    element.scrollIntoView()
+  }
+}
```
asouqi, line 18: we don't use this!

asouqi (changes requested):

mbartenev-atypon (approved):

src/components/Panel.tsx
mnatsheh: Removed.

src/components/track-changes/TrackChangesPanel.tsx
mnatsheh: Added.

src/components/track-changes/suggestion-list/Suggestion.tsx
mnatsheh: Added in `TrackChangesPanel`.

src/components/track-changes/suggestion-list/Suggestion.tsx
mnatsheh: No longer relevant.

asouqi (approved):

PR #455, opened 2024-05-20
LEAN-3624 Fix for inspector tab appearance

mbartenev-atypon:


mbartenev-atypon:
I propose to merge that without QA as its a very small change and it works fine locally

IslamJMomani (approved):

PR #467, opened 2024-05-28
Unexclude Project and Manuscript models from the saved models 

mbetamony:
I see this was recently changed, and by doing this we're deleting the project & manuscript models


asouqi:
> and by doing this we're deleting the project & manuscript models

the API still keeps `MPProject` if we don't send it! I can see that locally! , the only thing that will be missing is the manuscript

asouqi (approved):

mbartenev-atypon (approved):

PR #480, opened 2024-06-06
DEMO

mbetamony:


src/quarterback/api/document.ts
```diff
@@ -99,23 +98,23 @@ export const listenStepUpdates = (
...
+      data.clientIDs
+    ) {
+      dataListener(data.version, data.steps, data.clientIDs)
     }
   }
-
+  //host to your websocket server locally, but to manuscripts api in production
   listen(
-    `doc/${projectID}/manuscript/${manuscriptID}/listen`,
+    `ws://localhost:35553/api/v2/doc/${projectID}/manuscript/${manuscriptID}/listen`,
```
mbetamony, line 114: I'm still not quite sure how this would work if we sent this straight to the api :thinking:

mbetamony: Actually it should theoretically work fine

PR #482, opened 2024-06-10
LEAN-3524: change the token payload

mbetamony:


mbetamony:
Are you guys OK with the changes here?

mbartenev-atypon:
Can you please provide a little info on the purpose of that change? The description in the ticket is a bit generic too. Also do I get it right that the "id" is not going represent a userID anymore? So that's not just renaming?

mbetamony:
> Can you please provide a little info on the purpose of that change? The description in the ticket is a bit generic too. Also do I get it right that the "id" is not going represent a userID anymore? So that's not just renaming?

Yeah sorry i didn't provide more context here. It was just renaming, but it made sense to revert it back to userID again, its more clear that way. 
We removed some of the unnecessary token payload in the BE. This PR is just to make sure both payload types between the FE and the BE are matching

hhusban (approved):

PR #489, opened 2024-06-24
Scroll into bibliography item - LEAN-3563

IslamJMomani:


asouqi (approved):

src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -25,6 +26,21 @@ import { SnapshotsDropdown } from '../inspector/SnapshotsDropdown'
 import { SortByDropdown } from './SortByDropdown'
 import { SuggestionList } from './suggestion-list/SuggestionList'
```
mnatsheh, line 28: I'm not a big fan of doing the scrolling here. We should be able to achieve the same result with https://prosemirror.net/docs/ref/#view.EditorProps.handleScrollToSelection while keeping the logic contained within the body-editor.

IslamJMomani: I will take care of this in LEAN-3694. Thanks for your input!

PR #502, opened 2024-07-04
Update from SSE to WebSockets

mbetamony:


src/quarterback/api/document.ts
```diff
@@ -99,23 +101,38 @@ export const listenStepUpdates = (
...
+        })})`
+      )
+    }
+    if (
+      typeof data.version != 'undefined' &&
+      data.steps &&
+      Array.isArray(data.steps) &&
+      data.clientIDs
+    ) {
+      dataListener(data.version, data.steps, data.clientIDs)
```
mbetamony, line 127: Do you guys think we should move these changes to the [body-editor listener,](https://github.com/Atypon-OpenSource/manuscripts-body-editor/blob/0ed378e3b19375dae16755d89056b86cd7a6d1ae/src/useEditor.ts#L58) and handle the data properly there?

mbartenev-atypon: As I see it, body editor should get data already prepared for consumption without details of how we fetched that and typecasted it. FYI I've changed the way we check for the transform version here - https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/524/files

src/quarterback/api/methods.ts
```diff
@@ -9,11 +9,6 @@
  *
  * All portions of the code written by Atypon Systems LLC are Copyright (c) 2022 Atypon Systems LLC. All Rights Reserved.
  */
-import { getVersion } from '@manuscripts/transform'
-import {
-  EventSourceMessage,
-  fetchEventSource,
-} from '@microsoft/fetch-event-source'
```
mbartenev-atypon, line 16: Are we going to use SSE for something else? Steps listening was the only usage for that as far as I know.

src/quarterback/api/methods.ts
```diff
@@ -150,55 +156,75 @@ export function del<T>(
...
-    },
-    onclose() {
-      // if the server closes the connection unexpectedly, retry:
-      console.log('EventSource connection closed')
-    },
-    onerror(err) {
-      console.log('EventSource connection error: ' + err)
-    },
-  })
+  const reconnectDelay = 1500
```
mbartenev-atypon, line 166: I wonder why it's 1500? It seem to me to be a good balanced value but I'm curious how it was produced.

mbetamony: I just found it to be kind of  balanced during testing (1 second felt too fast), i tried multiple things, one of the options were to have an exponential backoff with some sort of limit, but it felt like if we want something like that we'll need some sort of a popup to the users to let them know we're attempting a reconnect. I'm not sure if you have any thoughts on this.

src/quarterback/api/document.ts
mbetamony: I'll keep this unresolved as a reminder to update the code once your PR is merged.

mbartenev-atypon: Thanks

mbartenev-atypon (approved):

src/quarterback/api/methods.ts
mbartenev-atypon: Ok, thanks for the info

src/quarterback/api/methods.ts
```diff
@@ -149,48 +145,73 @@ export function del<T>(
...
+  const removeListners = () => {
+    try {
+      ws.removeEventListener('open', onOpen)
+      ws.removeEventListener('message', listener)
+      ws.removeEventListener('close', onClose)
+      ws.removeEventListener('error', onError)
+    } catch (error) {
+      console.error('Error removing websocket listeners', error)
+    }
+  }
```
mnatsheh, line 179: This feels like an unnecessary function. We can just have the logic in `closeWebSocket`.

src/quarterback/api/methods.ts
```diff
@@ -149,48 +145,73 @@ export function del<T>(
...
+
+  const handleWebSocketEvents = () => {
+    if (!ws) {
+      return
+    }
+    ws.addEventListener('open', onOpen)
+    ws.addEventListener('message', listener)
+    ws.addEventListener('close', onClose)
+    ws.addEventListener('error', onError)
+  }
```
mnatsheh, line 204: This feels like an unnecessary function. We can just have the logic in `join`.

mbetamony: done.

src/quarterback/api/methods.ts
mbetamony: done.

mnatsheh (approved):

PR #517, opened 2024-07-15
LEAN-3772 LW-FY25-Q1-S4: The Citations Context Menu is Displayed when…

fwaisi:
… Clicking the Citations suggestion


src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -44,7 +44,7 @@ export const TrackChangesPanel: React.FC = () => {
     } else {
       tr.setSelection(NodeSelection.create(state.doc, suggestion.from))
     }
-
+    document.body.dataset.trackModal = 'true'
```
mbartenev-atypon, line 47: we need to use established ways of communcation between components instead of working around them. Here a couple lines above we are setting a selection with transaction. You can set additional meta on transaction and by doing that communicate any info you need to the body editor state. Also please consider that maybe to activate modal on selection is not too specific and it is, actually, the logic of activating modal that should be adjusted in the first place to be more limited and not the logic that just sets selection like here. Maybe modal should be activated only on click and not on selection? Or maybe we need to activate modal we need to explicitly specify so in the transactions meta?

mbartenev-atypon (changes requested):

src/components/track-changes/TrackChangesPanel.tsx
fwaisi: Using meta didn't worked for me, when I used getMeta in body Editor it didn't retrieve any meta.

I have added this workaround to handle multiple cases. 

For example if we used click event only, it will not fire the citation editor popup when the user insert a new citation.

mbartenev-atypon: I would that when we setSelection nothing should happen in terms of popup. If you click on it or setSelection and, for example, meta('showModal', true), then we should show the modal.

fwaisi: Do you mean to handle the selection using the click event and remove the select node listener

fwaisi: But FYI, send the meta from article to the body won't work

mbartenev-atypon: Not sure what was the problem with the transaction meta, but if you use getMeta directly in a node view, there'll be no meta as view don't have access to transactions, so you need to use a plugin that will take the meta out of a transaction and then the nodeView can read it from plugins state.

mbartenev-atypon: I don't see a reason for being able to dispatch transaction from article editor but not send meta. Besides we already do this in 2 places.
src/components/projects/EditorElement.tsx:187
src/tracked-models/creators.ts:169

fwaisi: reverted, I have used eventListenr to hande the context menu visibility and the selection only to show the insertion popup.

PR #518, opened 2024-07-16
LEAN-3773 Upgrade to React 18

fwaisi:


src/store/useStore.ts
```diff
@@ -9,45 +9,27 @@
...
 
 import deeperEqual from '../lib/deeper-equal'
 import { dispatch, GenericStore, state } from './Store'
 import { useGenericStore } from './StoreContext'
-// import { useSyncExternalStoreWithSelector } from 'use-sync-external-store/with-selector';
+
 type Selector<T> = (r: state) => state | T
+
 export const useStore = <T>(
-  selector?: Selector<T>
```
mbartenev-atypon, line 20: why it's not optional anymore?

src/store/useStore.ts
```diff
@@ -9,45 +9,27 @@
...
-  // @TODO - in react 18 these hooks usage will have to be replaced with this to work correctly with the concurrent rendering
-  //   const state = useSyncExternalStoreWithSelector(
-  //     store.subscribe,
-  //     store.getState(),
-  //     undefined,
-  //     selector,
-  //     deeperEqual
-  //   );
+  const init = selector ? () => selector(store.state!) : () => store.state
+  const state = useSyncExternalStoreWithSelector(
```
mbartenev-atypon, line 26: did you test if it works well in terms of rerenders and selectors updates?

mbartenev-atypon: good catch with this update though, thanks

src/store/useStore.ts
fwaisi: fixed

src/store/useStore.ts
fwaisi: yes I have done a quick test and didn't notice any thing wrong.

PR #528, opened 2024-07-22
LEAN-3786 Reducing rerendering

mbartenev-atypon:
Reducing rerenders and removing trackedModelMap. Some additional changes were needed for removing trackedModel map in TrackedModel and in SectionInspector as both still relied on json schema models.

PR #531, opened 2024-07-24
Removing json schema

mbartenev-atypon:


mbartenev-atypon:
There is a couple more things that we need to do before we can proceed with complete removal of models:

1. Change the way we create snapshot so it's actually done on backend, or make backend accept Prosemirror doc for snapshot instead of the models list
2. We should merge changes related to websocket update as it woulde preferable to restructure the QuarterbackSource and quarterback api in this PR, which will put websocket work too much out of date.

mbartenev-atypon:
Usage of json-schema becomes very limited only 12 types will be used from it after this change:
1. Project
2. Manuscript
3. SectionCategory
4. ObjectTypes
5. BibliographicDate
6. BibliographicName
7. UserProfile
8. ManuscriptTemplate
9. Bundle
10. CommentAnnotation
11. Model
12. manuscriptIDTypes

That is across 11 files (and 2 test files)

src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -16,72 +16,56 @@ import {
...
+      // @TODO - check if trackstate actually issues some updates and we listen to them
+      trackState: store.trackState,
+      selectedSuggestionID: store.selectedSuggestionID,
+    })
+  )
 
-    view?.focus()
-    view?.dispatch(tr.scrollIntoView())
-  }
+  const onSelect = useCallback((suggestion: TrackedChange) => {
```
mnatsheh, line 39: Can we call this `handleSelect` for consistency?

mnatsheh (approved):
This looks to me like a superset of https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/528 so I'll just review this.

mnatsheh:
> Change the way we create snapshot so it's actually done on backend, or make backend accept Prosemirror doc for snapshot instead of the models list
If you're talking about the doc/snapshot endpoint, it doesn't need any data to work; it uses the doc from the server.

mbartenev-atypon:
> > Change the way we create snapshot so it's actually done on backend, or make backend accept Prosemirror doc for snapshot instead of the models list
> > If you're talking about the doc/snapshot endpoint, it doesn't need any data to work; it uses the doc from the server.

I meant saving doc when when create snapshot. https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/hooks/use-handle-snapshot.ts#L49 
In this function we encode the doc into the models and then send the models to backend.

mbartenev-atypon:
> This looks to me like a superset of #528 so I'll just review this.

I was planning to merge the #528 first since this changes are a bit more extensive and I used the branch from there as this changes should be based on the ones in #528. But I don't mind going with just this PR

mnatsheh:
> > > Change the way we create snapshot so it's actually done on backend, or make backend accept Prosemirror doc for snapshot instead of the models list
> > > If you're talking about the doc/snapshot endpoint, it doesn't need any data to work; it uses the doc from the server.
> 
> I meant saving doc when when create snapshot. https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/hooks/use-handle-snapshot.ts#L49 In this function we encode the doc into the models and then send the models to backend.

Ah OK. We should not need this call at all when we eliminate json-schema.

src/components/track-changes/TrackChangesPanel.tsx
mbartenev-atypon: done

PR #533, opened 2024-07-29
LEAN-3788: LW-FY25-Q1-S5: Added backmatter section is still shown as …

fwaisi:
…enabled


src/components/projects/CategoryInput.tsx
```diff
@@ -59,12 +61,25 @@ export const CategoryInput: React.FC<{
   const options = useMemo(() => {
     const options: OptionType[] = []
     sectionCategories.map((cat) => {
+      // check if the category is part of the backmatter section, and its already present in the document
+      let isDisabled = false
+      if (
+        cat.groupIDs &&
+        isBackMatterSection(cat.groupIDs[0]) &&
+        document.querySelector(`[data-category="${cat._id}"]`)
```
mbartenev-atypon, line 69: Please check against the prosemirror doc instead. This will also be safer, since in case we'd make changes to the section attributes, it'd be reflected on the type and will not build.

mbartenev-atypon (changes requested):

src/lib/section-categories.ts
```diff
@@ -61,3 +62,19 @@ export const isUniquePresent = (
 ) => {
   return isUnique(cat._id) && Object.keys(existingCats).includes(cat._id)
 }
+
+export const isBackMatterSection = (groupId: string): boolean => {
```
mbartenev-atypon, line 66: probably no need to specify return type

src/lib/section-categories.ts
```diff
@@ -61,3 +62,19 @@ export const isUniquePresent = (
...
+  return groupId === 'MPSectionCategory:backmatter'
+}
+
+export const isBackMatterSectionCategoryExist = (parent: ManuscriptNode, category: string): boolean => {
```
mbartenev-atypon, line 70: probably no need to specify return type

mbartenev-atypon (approved):

src/lib/section-categories.ts
```diff
@@ -61,3 +62,19 @@ export const isUniquePresent = (
...
+  let found = false;
+  parent.descendants((node: ManuscriptNode) => {
+    if (node.attrs.category === category) {
+      found = true;
+      return false;
```
mbartenev-atypon, line 75: you could also return false if found == true to practically shot down the walk-through

PR #543, opened 2024-08-04
LEAN-2960:LW-FY24- Q2-S5: File extension is not showing for the trunc…

nghazzawi-atypon:
Trimming the file name to show the file extension and Add a tooltip to display full file names without cropping, as recommended in comment [#2](https://jira.atypon.com/browse/LEAN-2960?focusedCommentId=6255294#comment-6255294).


src/components/FileManager/FileName-utils.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Software distributed under the License is distributed on an “AS IS” basis, WITHOUT WARRANTY OF ANY KIND, either express or implied. See the License for the specific language governing rights and limitations under the License.
+ *
+ * The Original Code is manuscripts-frontend.
+ *
+ * The Original Developer is the Initial Developer. The Initial Developer of the Original Code is Atypon Systems LLC.
+ *
+ * All portions of the code written by Atypon Systems LLC are Copyright (c) 2024 Atypon Systems LLC. All Rights Reserved.
+ */
+export const trimFilename = (filename: string, maxLength: number) => {
```
mnatsheh, line 12: Do we need to have this in a new file?

nghazzawi-atypon: @mnatsheh 
I think it would be better to place this in a separate file as a utility, making it reusable in other parts of the product if needed. This way, we can easily access this functionality wherever it's required.

mnatsheh: I'm not really sure how useful this would be elsewhere, but if you prefer to have it separate, please move it to `lib/files.ts`.

nghazzawi-atypon: I was initially trying to follow to the standard like: footnotes. We have a footnote utility with the necessary functions, and I thought that truncating names might be used frequently in the editor. However, moving this functionality to `lib/files.ts` seems like a better approach. I'll make that change.

nghazzawi-atypon: done

mnatsheh (approved):

PR #559, opened 2024-08-14
LEAN-3811

asouqi:


src/components/comments/CommentBody.tsx
```diff
@@ -67,17 +73,21 @@ const EditorActions = styled(ButtonGroup)`
 export interface CommentBodyProps {
   comment: Comment
   isEditing: boolean
+  isNew: boolean
   onSave: (content: string) => void
   onCancel: () => void
   onSelect: () => void
+  onDelete: (id: string) => void
```
mnatsheh, line 80: I don't think we need to pass `isNew` and `onDelete` to this component. Calling `onCancel` should be enough, since the same logic was also added to `handleCancel` in the parent component.

asouqi: oh that is right, updated

mnatsheh (approved):

PR #561, opened 2024-08-15
Use config passed from the parent app - LEAN-3787

IslamJMomani:


src/config.ts
```diff
@@ -10,68 +10,26 @@
...
-  },
-  support: {
-    email: process.env.SUPPORT_EMAIL || 'support@manuscriptsapp.com',
-  },
-  quarterback: {
-    enabled: isTrue(process.env.QUARTERBACK_ENABLED),
-    url: normalizeURL(process.env.QUARTERBACK_URL) || '',
-  },
-  keywordsCategories: isTrue(process.env.FEATURE_KEYWORDS_CATEGORIES),
+let config: Config
```
mnatsheh, line 27: Can we store the config in react state or context?

src/config.ts
```diff
@@ -10,68 +10,26 @@
...
-const isTrue = (value: string | undefined) => {
-  return value === '1' || value === 'true'
-}
-
-const normalizeURL = (url: string | undefined) => {
-  return url && url.replace(/\/$/, '')
-}
-
-interface Config {
+export interface Config {
```
mnatsheh, line 13: Let's call this `ManuscriptsEditorConfig`. We should be able to use this in the parent app so to avoid duplicating the type definition.

src/index.tsx
```diff
@@ -34,13 +34,14 @@ export type {
   AppStateObserver,
 } from './EditorApp'
 
-const ManuscriptEditor: React.FC<EditorAppProps> = ({
+const ManuscriptEditor: React.FC<EditorAppProps & { config: Config }> = ({
```
mnatsheh, line 37: Can we just add the new prop to `EditorAppProps`?

src/index.tsx
```diff
@@ -61,6 +62,7 @@ const ManuscriptEditor: React.FC<EditorAppProps> = ({
     }
   }, [authToken])
 
+  setConfig(config)
```
mbartenev-atypon, line 65: can we move it into a useEffect? Especially if we are going to use context to pass it around.

mbartenev-atypon (changes requested):

src/config.ts
IslamJMomani: 2 reasons why I didn't use Context:

- I don’t expect the configuration to change, so using Context would introduce unnecessary complexity.
- Context cannot be directly used in class constructors or non-React code.

src/config.ts
IslamJMomani: Done

src/index.tsx
IslamJMomani: Given that I don't expect the config to change and to avoid the overhead of using `useEffect`, I updated the code to use `useMemo`. And regarding using Context, please check my comment above ☝️  and tell me what you think.

src/index.tsx
IslamJMomani: I could find other options to clean this up, but I didn’t want to add `config` to `EditorAppProps` to avoid the need to pass it to every component that uses `EditorAppProps`, like `EditorApp` and `Main`.

mnatsheh: Got it.

src/index.tsx
mbartenev-atypon: as things are right now, updating it will not cause any reactive updates and this memo will rerun anytime it's updated anyway, so if we'd like to keep the current approach, we probably wouldn't need it. My suggestion to move it into useEffect (sorry,  I should've specified - the one with the empty deps), was to allow setting config only on component mount. What would be the overhead of using useEffect here?

src/config.ts
IslamJMomani: Thank you, Mikhail, for bringing these points up. I agree that ensuring data immutability would be beneficial, and I hadn't considered that aspect. Your suggestion about relying on the React ecosystem to manage that and ensure immutability makes sense, but the idea that context cannot be directly used in non-React code will be very annoying, especially for the configuration case, and we will need to drill props along in many places (e.g., see `methods.ts`). Do you have any thoughts or suggestions on how to tackle that if we stick with using context?

IslamJMomani: Seems like I'll stick with the current implementation, but I'll update it to prevent changes to the config variable once it's set up for the first time (as you mentioned in your first point). Any objections?

src/index.tsx
IslamJMomani: Maybe you’re right, `useMemo` might not be suitable. However, regarding `useEffect`, even with an empty dependency array (which resolves the overhead concern), we will still face a timing issue because it runs after the component is rendered. And as you know, we need to set the configuration before rendering begins.

src/config.ts
mbartenev-atypon: Sounds good, thanks Islam.

src/index.tsx
mbartenev-atypon: Got it. Thanks for the explanation.

mbartenev-atypon (approved):

PR #569, opened 2024-08-25
LEAN-3892 FNS-130 - Inspector Track Changes for footnotes

fwaisi:


src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -7,61 +7,201 @@
...
-      </SnippetText>
-    </>
+    <SnippetText isRejected={dataTracked.status === CHANGE_STATUS.rejected}>
+      {snippet ? (
+        <>
+          <Operation color={dataTracked.operation}>
+            {snippet.operation}:
+          </Operation>
+          <NodeName>{snippet.nodeName}:</NodeName>
+          <Content>{parse(snippet.content)}</Content>
```
mbartenev-atypon, line 148: Why do we need to parse that html into react tree? Are we going to work with it as with active react dom tree? Can we just insert html into ref?

fwaisi: @mbartenev-atypon Why do we need to parse that html into react tree? Better for rendering and debugging.
Are we going to work with it as with active react dom tree? I don't think so.
Can we just insert html into ref? Yes, sure we can. But we have to ensure that the HTML content is sanitized, which I think we need it also for the current implantation.

I will keep this as note while implementing the rest nodes in the next sprint.

mbartenev-atypon: Why do you think we need to santize it? Why it this better for debugging? Also in most cases it should be just a piece of text. Introducing a new dependency to parse a small piece of html is bit of an overkill. Additionally, whatever are the design requirement here, we should adjust them, if needed, to allow us to display the result here in a controlled way and not just dropping in html from the editor.

fwaisi: Why do you think we need to sanities it?, because we are going to render whatever the user type, so if the user typed scripts it will be rendered.
Why it this better for debugging? it will be included in the react dom and tree.

Also in most cases it should be just a piece of text. Introducing a new dependency to parse a small piece of html is bit of an overkill. Additionally, whatever are the design requirement here, we should adjust them, if needed, to allow us to display the result here in a controlled way and not just dropping in html from the editor.
Totally agree, I was planing to do it by ref or css because its only needed for the alphabet order to match the style, so I will try to find a better way to render it in the next sprint without introducing any extra dependencies.

mbartenev-atypon (approved):

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -7,61 +7,201 @@
...
+          return {
+            snippet: {
+              operation: changeOperationAlias(dataTracked.operation),
+              nodeName: suggestion.node.type.spec.name,
+              content: getInlineFootnoteContent(doc, suggestion.attrs),
+            },
+            message: '',
+          }
+        } else if (suggestion.node.type === schema.nodes.footnote) {
+          console.log('suggestion.node', suggestion.node, suggestion)
```
IslamJMomani, line 80: @fwaisi, Did you leave this intentionally?

fwaisi: No, good catch.

Will remove it shortly

src/lib/tracking.ts
```diff
@@ -40,3 +40,20 @@ export const trackEvent: TrackEvent = ({ category, action, label, value }) => {
     })
   }
 }
+
+export const changeOperationAlias = (operation: string): string => {
```
mbartenev-atypon, line 44: @fwaisi , since you are still working on that, please use enum CHANGE_OPERATION for this

PR #570, opened 2024-08-26
LEAN-3902 Update @reach/tabs dependency

fwaisi:


package.json
```diff
@@ -32,6 +32,9 @@
     "version": "ALLOW_MISSING_VARIABLES=1 yarn build"
   },
   "dependencies": {
+    "@chakra-ui/react": "^2.8.2",
+    "@emotion/react": "^11.13.3",
+    "@emotion/styled": "^11.13.0",
```
mnatsheh, line 37: Why are we adding these extra dependencies?

fwaisi: its a dependency for the chakra/ui

package.json
```diff
@@ -32,21 +32,24 @@
     "version": "ALLOW_MISSING_VARIABLES=1 yarn build"
   },
   "dependencies": {
+    "@chakra-ui/react": "^2.8.2",
```
mnatsheh, line 35: Can we depend only on the tabs component (@chakra-ui/tabs)?

fwaisi: No, we can't rely on it alone.

After updating the import from the package, we're now only consuming 5KB of it.

package.json
mbartenev-atypon: I think we should use headless ui, because it doesn't require any peer deps except for react and react-dom, which is fair. It also requires a bit fewer dependencies in total and is considerably more popular. I would very much prefer not to have peer dependencies that are not explicitly used in the project.

mbartenev-atypon (changes requested):

fwaisi:
Closing, since the branch is outdated and I have pushed new implementation in the following branch LEAN-3902-1

PR #594, opened 2024-09-12
DataSource cleanup

mnatsheh:
A bunch of small cleanups centered about the store data sources:
- Replace postgres/ps with api.
- Remove the special quarterback code, and merge it with api. 

There is also a couple of cleanups related to the quarterback piece:
- Simplified the collabprovider implementation a bit.
- Changed what "quarterback" data is maintained in the store, and simplified how snapshots are handled.


src/api/types.ts
```diff
@@ -7,24 +7,39 @@
...
-  role: string,
-  message = 'message'
-) =>
-  client.post(`/invitation/${encodeURIComponent(projectID)}/invite`, {
-    invitedUsers,
-    role,
-    message,
-  })
+export type SendStepsPayload = {
+  steps: unknown[]
```
IslamJMomani, line 20: Why steps type defined as unknown?

src/api/buildData.ts
```diff
@@ -80,27 +96,47 @@ const getUserData = async (projectID: string, user: UserProfile, api: Api) => {
...
   const role = project ? getUserRole(project, user.userID) : null
   const users = await getUserData(projectID, user, api)
 
+  const stepsExchanger = new StepsExchanger(
+    projectID,
+    manuscriptID,
+    doc.version,
+    api
+  )
+  stepsExchanger.isThrottling.onChange((value: boolean) => {
```
mbartenev-atypon, line 119: Do you think we can make it work without putting stepExchanger into the store? This looks like something what shouldn't be generally accessible to components since it's designed specifically for editor props and works directly with api. When I created it, I wanted to have no reference to it in the article editor on purpose to avoid potential misuse.

src/lib/doc.ts
```diff
@@ -7,12 +7,34 @@
...
  * The Original Developer is the Initial Developer. The Initial Developer of the Original Code is Atypon Systems LLC.
  *
- * All portions of the code written by Atypon Systems LLC are Copyright (c) 2022 Atypon Systems LLC. All Rights Reserved.
+ * All portions of the code written by Atypon Systems LLC are Copyright (c) 2024 Atypon Systems LLC. All Rights Reserved.
  */
 import { CHANGE_OPERATION } from '@manuscripts/track-changes-plugin'
 import { ManuscriptEditorState, ManuscriptNode } from '@manuscripts/transform'
 
-export function filterUnchangedContent(node: ManuscriptNode) {
+export type PMDoc = any
```
mbartenev-atypon, line 15: Can we improve on that? I think there is a JSONNode type that we use in transform, that probably can be used here instead of any or we can use just some generic JSON type too.

src/api/StepsExchanger.ts
```diff
@@ -0,0 +1,135 @@
...
+    if (!flush) {
+      this.isThrottling.setValue(true)
+    }
+
+    return Promise.resolve()
+  }
+
+  async receiveSteps(version: number, steps: unknown[], clientIDs: number[]) {
+    if (steps.length) {
+      //TODO send steps to listener
```
mbartenev-atypon, line 108: we can probably remove that comment

mbartenev-atypon (changes requested):

src/api/types.ts
mnatsheh: This is how it was before. I updated it now to be `Steps[]`.

src/api/StepsExchanger.ts
mnatsheh: This is a comment I added so we don't forget to do this. The current body-editor api doesn't accept any args, and I didn't want to change it now.

src/lib/doc.ts
mnatsheh: The old quarterback code had a `JsonValue` type imported from the `type-fest`  package, but I think it's overkill to have that package just for this type. I can add it back if it's preferred.

mnatsheh: Alternativally, I can define a general type for JSON and use it.

src/api/buildData.ts
mnatsheh: I can do that, but it would require storing the version to the store, and somehow get access to the api. I can just create a new `Api` instance if that's acceptable.

src/lib/doc.ts
mbartenev-atypon: I was thinking that we can use the type from transform - 
[export type JSONNode ](https://github.com/Atypon-OpenSource/manuscripts-transform/blob/master/src/schema/migration/migrate.ts#L23), but anything that at least broadly suggest a document JSON-like structure will do just as well.

src/api/buildData.ts
```diff
@@ -80,27 +96,47 @@ const getUserData = async (projectID: string, user: UserProfile, api: Api) => {
...
 
-  const manuscriptNode = doc as ActualManuscriptNode
-  const state = await getManuscriptData(api, manuscriptNode.attrs.prototype)
+  const doc = await getDocumentData(projectID, manuscriptID, api)
+  const state = await getManuscriptData(doc.doc.attrs.prototype, api)
   const [project, manuscript] = await getIdModels(api, projectID, manuscriptID)
   const role = project ? getUserRole(project, user.userID) : null
   const users = await getUserData(projectID, user, api)
 
+  const stepsExchanger = new StepsExchanger(
```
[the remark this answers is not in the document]
mbartenev-atypon: Ok

src/api/buildData.ts
mbartenev-atypon: I think we can either create a new api instance (we can singleton it if it's not safe for WS listeners) but I'm not sure why we need to store the current version in the store. As far as I remember the only reason we need to set the current version is to initiate collab plugin on when we load/reload the editor and that initial version value you already add to the store by unpacking output of getDocumentData into the store on line 136 in this file (probably would be better to put it into the store with a label like 'initialVersion' or similar).

src/api/StepsExchanger.ts
```diff
@@ -0,0 +1,133 @@
...
+  isThrottling: ObservableBoolean
+  attempt = 0
+
+  constructor(
+    projectID: string,
+    manuscriptID: string,
+    currentVersion: number,
+    api: Api
+  ) {
+    super()
```
mbartenev-atypon, line 55: In the original implementation there was a check to prevent duplicated instantiation of this class. I would suggest to keep it as there is sometimes little control of what and how react reloads components (especially with upgrades to newer versions). Additional complications may occur in development when react runs each hook twice and hot reload is active.

src/api/buildData.ts
mnatsheh: I made more changes and got the stepsExchanger out of the source and created it in `useCreateEditor` instead.

src/lib/doc.ts
mnatsheh: That type is defined under "migration" so I prefer not to use it in its current state. I would rather not update the transform for now (as it would result in several other components being published), so I can move the type and use it here after this is merged if you don't mind.

mbartenev-atypon: Ok. Sounds good

src/api/StepsExchanger.ts
mbartenev-atypon: I see. I wasn't aware of that.

mbartenev-atypon (approved):

PR #602, opened 2024-09-22
LEAN-3958 EI25Q1-010 - Implement texts for Track changes

fwaisi:


src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -42,115 +48,152 @@ export const SuggestionSnippet: React.FC<{ suggestion: TrackedChange }> = ({
     doc: store.doc,
   }))
   const [snippet, setSnippet] = useState<SnippetData | null>(null)
-  const [message, setMessage] = useState('')
   const { dataTracked } = suggestion
+  const contentRef = useRef<HTMLDivElement>(null)
+  const isContainLaTeX = (text: string): boolean => {
```
mbartenev-atypon, line 53: this probably doesn't need to be defined in the context of react component and can be move to lib

src/lib/fonts.ts
```diff
@@ -13,3 +13,7 @@
 import '@fontsource/lato'
 import '@fontsource/pt-sans'
 import '@fontsource/pt-serif'
+// Importing specific weights
```
mbartenev-atypon, line 16: I would say that to me this quite clear what's going on here with a comment

src/lib/utils.ts
```diff
@@ -33,6 +34,78 @@ export const getNodeTextContent = (node: ManuscriptNode) => {
...
+    'text/html'
+  ).body.textContent
+  return textContent ? textContent : ''
+}
+
+export const getFigureLabel = (
+  state: ManuscriptEditorState,
+  node: ManuscriptNode
+) => {
+  const objectsPlugin = findPluginByKey(state, 'objects')
```
mbartenev-atypon, line 103: While do sometimes refer to plugins directly by string in the body-editor, it would be advisable to use a constant or some other reference imported from body-editor to address the plugin here to avoid bugs if a plugin gets renamed or refactored.

src/lib/utils.ts
```diff
@@ -33,6 +34,78 @@ export const getNodeTextContent = (node: ManuscriptNode) => {
...
+) => {
+  let textContent = ''
+  textContent += `${node.attrs.institution}`
+  if (oldAttrs) {
+    textContent = `${oldAttrs.institution} to ` + textContent
+  }
+  return textContent
+}
+
+export const findBibliographyById = (
```
mbartenev-atypon, line 63: Can you point me to where this is used? I can't find any usage.

src/lib/utils.ts
```diff
@@ -33,6 +34,78 @@ export const getNodeTextContent = (node: ManuscriptNode) => {
   return textContent
 }
 
+export const getContributorTextContent = (
```
mbartenev-atypon, line 37: Since we create all these functions for a similar purpose and they do same things, I suggest, we organise them accordingly: they can be moved into a common for them file or directory. Utils.ts is really more for assorted items that are shared between different components. 

Another neat thing that can be done here is to refactor those get{node}TextContent function in such a way that content getters are accessible by schema.nodeType.

Perhaps we need to consider moving those textContentGetters into transform since the can be directly associated with each nodeSchema.

mbartenev-atypon (changes requested):

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
fwaisi: I have changed the implementation

src/lib/utils.ts
fwaisi: done

src/lib/utils.ts
fwaisi: deleted

src/lib/utils.ts
fwaisi: > Since we create all these functions for a similar purpose and they do same things, I suggest, we organise them accordingly: they can be moved into a common for them file or directory. Utils.ts is really more for assorted items that are shared between different components.
> 
sure, will do.

fwaisi: > Another neat thing that can be done here is to refactor those get{node}TextContent function in such a way that content getters are accessible by schema.nodeType.
> 


are you suggesting to refactor the textContent getter, to return the whole content for the target node and its children nodes?

> Perhaps we need to consider moving those textContentGetters into transform since the can be directly associated with each nodeSchema.

Something like have a function inside the schema to return the text content of the node?

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -42,115 +50,178 @@ export const SuggestionSnippet: React.FC<{ suggestion: TrackedChange }> = ({
...
+      contentRef.current.innerHTML =
+        purify.sanitize(': ' + snippet.content) || ''
+      if (snippet.isEquation) {
+        renderMath(contentRef.current)
+      }
+    }
+  }, [snippet?.content, snippet?.isEquation])
 
   useEffect(() => {
     const getSnippetData = (): {
```
mbartenev-atypon, line 66: looks like this function can accept arguments that it needs to produce a result instead of how it's setup right now and then it can be moved into a different file to make this component more readable.

src/lib/utils.ts
mbartenev-atypon: What I meant is something like a function that would get text for a node type that can be displayed in the inspector. So instead of creating a bunch of functions like getTextContentFor{some}Node, you can call getDisplayTextFor<NodeTypes.Affiliation>(this.node) (this is just an example)

mbartenev-atypon: So this is not a refactoring but a create of a function that would return a displayable text for the inspector

src/lib/track-changes.ts
```diff
@@ -0,0 +1,165 @@
+/*!
```
mbartenev-atypon, line 1: A lot of these functions are just to get displayable text for inspector. And they need to be grouped together based on this trait and not just in track-changes. For example getContributorTextContent doesn't have anything to do with track changes, but just creates displayable text for the node to be shown in the inspector. As I mentioned earlier you can create a single function that does that. Or as an alternative you can create a class that handles creation of displayable text for different types of nodes.

mbartenev-atypon (changes requested):

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -42,115 +50,178 @@ export const SuggestionSnippet: React.FC<{ suggestion: TrackedChange }> = ({
...
-              .charAt(0)
-              .toUpperCase()}${suggestion.node.type.name.slice(1)} ${
-              dataTracked.operation
-            }`,
+          case schema.nodes.list: {
+            return {
+              snippet: {
+                operation: changeOperationAlias(dataTracked.operation),
+                nodeName: suggestion.node.type.spec.name,
+                content: `<span class="inspector_listItem">${getFirstChildContent(
```
mbartenev-atypon, line 186: Please adjust this class name so it corresponds to our general approach. I think we just use dashes styles, so it'll be inspector-list-item.

mbartenev-atypon (changes requested):

src/lib/utils.ts
fwaisi: done

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
fwaisi: done

src/lib/change-handlers.ts
```diff
@@ -0,0 +1,160 @@
...
+  view: any,
+  doc: any,
+  dataTracked: any
+): SnippetData | null => {
+  const nodeContentRetriever = new NodeTextContentRetriever(view.state)
+
+  switch (suggestion.node.type) {
+    case schema.nodes.inline_footnote: {
+      return {
+        operation: changeOperationAlias(dataTracked.operation),
```
IslamJMomani, line 60: Destructure `nodeName` and `operation` for cleaner code and to avoid repetitive access. Add these lines just before the switch statement:
```
const { node } = suggestion
const operation = changeOperationAlias(dataTracked.operation)
const nodeName = node.type.spec.name || node.type.name
```

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -10,28 +10,26 @@
...
+  handleUnknownChange,
+} from '../../../lib/change-handlers'
 import { useStore } from '../../../store'
+import SnippetContent from './SnippetContent'
 
 interface SnippetData {
   operation: string
   nodeName: string
   content: string
+  isEquation?: boolean
```
IslamJMomani, line 32: Can't you just use `nodeName` to check if it is an equation?

src/lib/change-handlers.ts
fwaisi: done

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
fwaisi: Yes, it's possible. However, implementing it this way will require more logic and extra code. I believe the current approach is much shorter and serves the purpose effectively.

src/lib/change-handlers.ts
IslamJMomani: Just pass `operation` and `nodeName` directly; there's no need for `operation: operation` and `nodeName: nodeName`. Also, I don't see you using `node` instead of `suggestion.node`.

fwaisi: done

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
fwaisi: done

IslamJMomani (approved):

IslamJMomani (approved):

mbartenev-atypon (approved):

PR #622, opened 2024-10-08
Replying to comments - LEAN-3894

IslamJMomani:
Given the design: https://www.figma.com/design/h8mdcWWaCnFtWe7OGtLq1y/Reply-to-comments?node-id=18923-8987&t=TI8Rlv6pheZdgsiu-1

- Added the ability to reply to a comment.
- Grouped comments and their replies into threads.
- Added confirmation for comment deletion.
- Implemented expand/collapse functionality for threads.


src/components/comments/CommentActions.tsx
```diff
@@ -69,13 +74,17 @@ export const CommentActions: React.FC<CommentActionsProps> = ({
   const { isOpen, toggleOpen, wrapperRef } = useDropdown()
 
   return (
-    <>
+    <Container>
       {isResolveEnabled && (
         <CommentResolveButton comment={comment} onClick={toggleResolve} />
       )}
       {isActionsEnabled && (
         <DropdownContainer ref={wrapperRef}>
```
asouqi, line 82: should the `DropdownList` be smaller similar to the design

asouqi:
@IslamJMomani should we allow the deletion of comments with a list of replies, instead of deleting replies one by one!

src/components/comments/ReplyBox.tsx
```diff
@@ -0,0 +1,129 @@
...
+  ButtonGroup,
+  PrimaryButton,
+  SecondaryButton,
+} from '@manuscripts/style-guide'
+import React, { ChangeEvent, useMemo, useRef, useState } from 'react'
+import styled from 'styled-components'
+
+interface ReplyBoxProps {
+  insertCommentReply: (target: string, contents: string) => void
+  commentId: string
```
mnatsheh, line 23: Let's call this `commentID`.

mnatsheh (approved):

IslamJMomani:
> @IslamJMomani should we allow the deletion of comments with a list of replies, instead of deleting replies one by one!

I discussed this before with Panos and Vip. I suggested following what Google Docs does, which is allowing editing/deletion for any comment or reply. However, they insist they don't want to allow deletion or editing of a comment or reply unless it is the last one.

src/components/comments/ReplyBox.tsx
IslamJMomani: Done

src/components/comments/CommentActions.tsx
IslamJMomani: Done

asouqi (approved):

PR #642, opened 2024-11-03
LEAN-4090 LW-FY25-Q2: Error 'Cannot read properties of undefined (rea…

fwaisi:
…ding 'state')' on Completing Task with Pending Suggestions


IslamJMomani (approved):
Can you change it to:

`
if (!view) {
  return null
}
`

to return early for better readability?

PR #651, opened 2024-11-10
LEAN-4073: LW-FY25-Q2-rollover: Scrolling is not working correctly af…

nghazzawi-atypon:
the scroll to target comment is not working in view mode (edit article permissions). I addressed this by implementing a manual scroll using the existing scrollIntoView function defined on line 45


mnatsheh:
I don't like this approach. This looks like a PM problem, so I'll post a question for them. In the meantime, we can handle this in `handleScrollToSelection` in the editor, which currently has logic for bib items so we can extend it.

nghazzawi-atypon:
> I don't like this approach. This looks like a PM problem, so I'll post a question for them. In the meantime, we can handle this in `handleScrollToSelection` in the editor, which currently has logic for bib items so we can extend it.

I have updated the PR, now I am extending the `handleScrollToSelectedElement` function (previously `handleScrollToBibliographyItem` from body editor) on this PR : https://github.com/Atypon-OpenSource/manuscripts-body-editor/pull/452 
to support various HTML elements, such as comments. The function has been re-implemented to pass other HTML elements rather than only bibliography items.

mnatsheh:
@nghazzawi-atypon , do we actually need to do any change here? can't we just have the body-editor handle the scrolling based on the selection?

PR #655, opened 2024-11-17
Toggle track changes styles On/Off

IslamJMomani:


src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -75,10 +87,22 @@ const ManuscriptPageView: React.FC = () => {
...
+                      </IconButton>
+                    </>
+                  )}
                 </ManuscriptMenusContainer>
                 {can.seeEditorToolbar && <ManuscriptToolbar />}
               </EditorHeader>
-              <EditorBody className="editor-body">
+              <EditorBody
+                className={`editor-body ${
+                  isTrackingEnabled && 'tracking-enabled'
```
mbartenev-atypon, line 103: Since all the changes related to toggling this class happen inside the body editor, it seems to me, that the body-editor should toggle this class inside itself instead of article-editor controlling it from outside and potentially there can be other logic inside body editor that will depend on tracking being visible or hidden.

src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -62,10 +65,19 @@ const ManuscriptPageContainer: React.FC = () => {
 
 const ManuscriptPageView: React.FC = () => {
   const can = usePermissions()
+  const [isTrackingOn, setTrackingOn] = useState(true)
+
+  const toggleTracking = () => {
+    setTrackingOn(!isTrackingOn)
```
mbartenev-atypon, line 71: If I understand correctly, here we just toggle suggestions visibility and not really disable track changes. Taking in account that we actually have a case and logic where we indeed disable TC completely it may be a good idea to consider using a more specific naming here to avoid confusion (i mean in general in this function). Maybe something like tracking-visible/tracking-hidden.

src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -62,10 +65,19 @@ const ManuscriptPageContainer: React.FC = () => {
 
 const ManuscriptPageView: React.FC = () => {
   const can = usePermissions()
+  const [isTrackingOn, setTrackingOn] = useState(true)
```
mbartenev-atypon, line 68: Can you please move the functionality introduced by this PR into a custom hook and call it here if possible? This is because ManuscriptPageView is a relatively high level component and if we can abstract away an implementation of a feature, that could help keep it cleaner. What do you think?

mbartenev-atypon (changes requested):

src/components/projects/ManuscriptPageContainer.tsx
IslamJMomani: Done

src/components/projects/ManuscriptPageContainer.tsx
IslamJMomani: Makes sense. Updated!

src/components/projects/ManuscriptPageContainer.tsx
IslamJMomani: I'm not sure what you are actually suggesting. Could you please elaborate?

Toggling lives in the article editor, and in the last commit, I moved the class name `tracking-visible` one level up so it's visible for all components.

mbartenev-atypon: What I meant is maybe we need to toggle that className inside the body editor since the css for this class is inside the body editor

mbartenev-atypon (approved):

src/components/projects/ManuscriptPageContainer.tsx
IslamJMomani: In my opinion, having the className at a higher level is more beneficial, especially in scenarios where multiple child components need to access it, not only in the body but also in the sidebar or inspector. This approach seems simpler and reduces redundancy.

I will proceed with the current behavior as you suggested, and of course, we can change that if needed, but I don't see a need for it right now.

PR #656, opened 2024-11-17
LEAN-4100 Notify WM about title changes

mbartenev-atypon:


src/lib/get-node-without-tc.ts
```diff
@@ -0,0 +1,66 @@
...
+
+export function checkSuggestionsFor(
+  node: JSONNode,
+  onNodeReal: (node: JSONNode) => void
+) {
+  let textString = ''
+  if (node.content?.length) {
+    const newContent = []
+
+    nodesLoop: for (let i = 0; i < node.content?.length; i++) {
```
asouqi, line 24: can we use node descendants, and build text when going through title node

asouqi (changes requested):

src/hooks/use-watch-title.ts
```diff
@@ -0,0 +1,48 @@
...
+        newTitleNode = node
+      }
+    })
+
+    if (newTitleNode) {
+      const node = newTitleNode as ManuscriptNode
+
+      const newRealText = getNodeRealText(node)
+
+      if (newRealText !== prevText) {
```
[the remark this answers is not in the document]
mbartenev-atypon: good point, thanks! changed to using the store

src/lib/get-node-without-tc.ts
mbartenev-atypon: I'm using this approach here because it allows not only to get textContent without TC but to recursively get nodes without track changes, which is a more generic case we will need for other tasks. If'll use descendants I can only retrieve text, since I cannot modify the nodes as it will affect the actual document.

asouqi (approved):

PR #671, opened 2024-11-27
LEAN-4102

asouqi:


src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -46,6 +46,16 @@ export const TrackChangesPanel: React.FC = () => {
 
   const { changeSet } = trackState || {}
 
+  const changes = useMemo(
+    () =>
+      changeSet?.groupChanges.filter(
+        (c) =>
+          c.dataTracked.status === CHANGE_STATUS.pending &&
+          c.dataTracked.operation !== CHANGE_OPERATION.reference
```
mbartenev-atypon, line 54: I would say that these details should be covered inside of ChangeSet and hidden by the abstraction that I suggested earlier. We may have a lot of different types of changes and this UI component should reason about how which one are displayable and which are groups and which are not. 
I would suggest that when we do "changeSet?.pending" it will already filter away all the changes that shouldn't be use here or in any other UI component.

asouqi: updated

src/components/track-changes/TrackChangesPanel.tsx
```diff
@@ -37,22 +36,27 @@ export const TrackChangesPanel: React.FC = () => {
...
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
   const execCmd = useExecCmd()
 
   const { changeSet } = trackState || {}
 
-  const handleAccept = useCallback((change: TrackedChange) => {
-    setChangeStatus(change, CHANGE_STATUS.accepted, execCmd)
+  const changes = useMemo(
```
mbartenev-atypon, line 48: can you please explain why do we need memo here? thanks

src/components/track-changes/suggestion-list/Suggestion.tsx
```diff
@@ -26,15 +26,15 @@ const scrollIntoView = (element: HTMLElement) => {
 }
 
 interface Props {
-  suggestion: TrackedChange
+  suggestions: TrackedChange[]
```
mbartenev-atypon, line 29: as basically and TrackedChange now represented by an array, it may be a bit more semantically precise to include the array inside the type. i.e. to rename "TrackedChange" for example to be "InternalChange", and now use the name for TrackedChange = Array<InternalChange>, does it make sense?

mbartenev-atypon (approved):

src/components/track-changes/TrackChangesPanel.tsx
asouqi: removed

src/components/track-changes/suggestion-list/Suggestion.tsx
asouqi: updated to the new type

mbartenev-atypon (approved):

PR #762, opened 2025-03-10
Lean 4355

fwaisi:


src/components/FileManager/FileActions.tsx
```diff
@@ -31,19 +32,35 @@ export const FileActions: React.FC<{
...
   const { isOpen, toggleOpen, wrapperRef } = useDropdown()
   const [isMoveDialogOpen, setMoveDialogOpen] = useState<boolean>(false)
+  const [isUseAsMainDialogOpen, setUseAsMainDialogOpen] =
+    useState<boolean>(false)
 
   const showDownload = can?.downloadFiles && onDownload
   const showReplace = can?.replaceFile && onReplace
   const showDetach = can?.detachFile && onDetach
   const showMove = can?.moveFile && move
+  const showUseAsMain =
```
mnatsheh, line 57: Do we need the section type check? Let's also check the `can.moveFile` permission.

src/components/FileManager/OtherFilesSection.tsx
```diff
@@ -66,16 +69,36 @@ export const OtherFilesSection: React.FC<{
...
   }
 
+  const asMainDocument = async (file: FileAttachment) => {
+    const mainDocument = findChildrenByType(
+      view.state.doc,
+      schema.nodes.attachment
+    )[0]
+    const tr = view.state.tr
+    tr.setNodeAttribute(mainDocument.pos, 'href', file.id)
+    view.dispatch(skipTracking(tr))
```
mnatsheh, line 79: This will not work if there is no existing main doc. We should just call `insertAttachment` and have it handle the uniqueness.

src/components/FileManager/SupplementsSection.tsx
```diff
@@ -88,9 +90,31 @@ export const SupplementsSection: React.FC<SupplementsSectionProps> = ({
...
 
+  const handleUseAsMain = async (supplement: NodeFile) => {
+    const mainDocument = findChildrenByType(
+      view.state.doc,
+      schema.nodes.attachment
+    )[0]
+    const tr = view.state.tr
+    const from = supplement.pos
+    const to = from + supplement.node.nodeSize
+    tr.setNodeAttribute(mainDocument.pos, 'href', supplement.file.id)
```
mnatsheh, line 101: Same as above.

src/components/FileManager/MainFilesSection.tsx
```diff
@@ -0,0 +1,199 @@
...
+          />
+        </MainDocumentContainer>
+      ) : (
+        <MainDocumentTitle>Upload main document</MainDocumentTitle>
+      )}
+      {can?.uploadFile && (
+        <FileUploader
+          onUpload={handleUpload}
+          placeholder="Drag or click to replace with a new file"
+          accept=".docx, .doc, .pdf, .xml"
```
mnatsheh, line 161: This check only applies when the file is uploaded here, but not when we "move" a file from another tab. We should either remove this or add the check there as well.

mnatsheh (changes requested):

src/components/FileManager/FileActions.tsx
fwaisi: Do we need the section type check? no.

Done

src/components/FileManager/MainFilesSection.tsx
fwaisi: I have added the check in the fileactions file

src/components/FileManager/SupplementsSection.tsx
fwaisi: done

mnatsheh (approved):

PR #801, opened 2025-04-22
LEAN-4475: LW-FY25-Q4: The lable of the track changes changes from "Title type" to a "Text" once adding a text to title field

nghazzawi-atypon:


src/lib/change-handlers.ts
```diff
@@ -38,6 +39,25 @@ interface SnippetData {
...
+  if (isAltTitleNode(node)) {
+    const type = node.attrs.type
+    switch (type) {
+      case 'running':
+      case 'right-running':
+        return 'Running Title'
+      case 'short':
+        return 'Short Title'
+      default:
+        return `${type.charAt(0).toUpperCase() + type.slice(1)} Title`
```
mbartenev-atypon, line 55: maybe I'm missing something, but wouldn't default case works for all the case except Right Running? And if so, we can just split the string on  dash ("-") character to process all of them?

mbartenev-atypon (approved):

src/lib/change-handlers.ts
```diff
@@ -38,6 +39,22 @@ interface SnippetData {
...
 }
 
+const isAltTitleNode = (node: ManuscriptNode): boolean =>
+  node.type === schema.nodes.alt_title
+
+const getTitleDisplayName = (node: ManuscriptNode): string => {
+  if (isAltTitleNode(node)) {
+    const type = node.attrs.type
+    return type === 'short'
+      ? 'Short Title'
```
mbartenev-atypon, line 49: will the default case not work for "short"?

nghazzawi-atypon: @mbartenev-atypon 
Committing test changes here to validate this on the dev environment, as I’m currently unable to test it locally. I’ll also test the case you mentioned for short titles and follow up with the final changes for review.

src/lib/change-handlers.ts
nghazzawi-atypon: done

src/lib/change-handlers.ts
nghazzawi-atypon: you are absolutely right ! .. updated

PR #810, opened 2025-04-24
File inspector when in an older snapshot - LEAN-2394

IslamJMomani:


src/components/projects/ManuscriptPageContainer.tsx
```diff
@@ -44,18 +44,22 @@ import { ManuscriptToolbar } from './ManuscriptToolbar'
...
-      project: state.project,
-      user: state.user,
-      permittedActions: state.permittedActions,
+  const [{ project, user, permittedActions, isViewingMode }] = useStore(
+    (state) => {
+      return {
+        project: state.project,
+        user: state.user,
+        permittedActions: state.permittedActions,
+        isViewingMode: state.trackState?.status === 'view-snapshots',
```
mbartenev-atypon, line 53: I think we should just have isViewingMode prop in the store instead of deriving it from trackState status. ops-editor and internal components can then just read it from the store. Potentially there may be other reasons why we might want to enable viewing mode.

src/hooks/use-create-editor.ts
```diff
@@ -78,8 +80,16 @@ export const useCreateEditor = () => {
     return () => stepsExchanger.stop()
   }, [dispatch, stepsExchanger])
 
-  const getCapabilities = memoize((project, user, permittedActions) =>
```
mbartenev-atypon, line 81: I wonder why we used memoize here before.

mbartenev-atypon (changes requested):

src/components/projects/ManuscriptPageContainer.tsx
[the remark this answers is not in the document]
IslamJMomani: Done

src/hooks/use-create-editor.ts
IslamJMomani: I'm not sure, to be honest.

mbartenev-atypon (approved):

PR #814, opened 2025-04-27
LEAN-4471: add expiry to tokens

mbetamony:
I've added expiry to tokens, when the token expires, we use an async function that refetches the token from WM. I found multiple token implementations inside this codebase, so i've removed most of it. I'm not sure if we can have a better implementation, or if you think we need dedicated FE work for this to better implement it.

Related work: 
https://github.com/atypon/leanworkflow-ops/pull/764
https://github.com/Atypon-OpenSource/manuscripts-api/pull/249
https://github.com/atypon/leanworkflow-manuscripts-integration/pull/51


src/api/Api.ts
```diff
@@ -33,13 +38,28 @@ import {
...
     })
+    this.getAuthToken = getAuthToken
+    this.instance.interceptors.request.use(this.authInterceptor)
+  }
+
+  authInterceptor = async (config: InternalAxiosRequestConfig) => {
+    if (TokenHandler.isExpired()) {
+      const token = await this.getAuthToken()
+      if (!token) {
+        throw new Error('failed to generate manuscripts token')
```
mbartenev-atypon, line 57: @mnatsheh , if we are unable to generate a new token, we probably need to somehow walk the user through that situation, for example redirect to a login page. Would that be appropriate approach, what do you think?

src/components/Page.tsx
```diff
@@ -48,13 +48,8 @@ const PageContainer = styled.div`
...
 
 export const Page: React.FC<{ children: React.ReactNode }> = ({ children }) => {
-  const [{ tokenData }] = useStore((store) => ({
-    tokenData: store.tokenData,
-  }))
-
-  if (!tokenData) {
-    throw new Error('TokenData not found. User is not loggged in!')
+  if (!TokenHandler.get()) {
+    throw new Error('Authentication to manuscripts required.')
```
mbartenev-atypon, line 52: I think we need a more generic manage, as manuscripts are normally embedded into another app. Probably something like 'Authentication to editor required.' We also have a page here for when the user is logged in but has no access to the document. We can probably reuse it and add a button there to redirect user to login. Alternatively we can trigger the reload of the page and it would normally trigger an appropriate redirect to Connect.

src/EditorApp.tsx
```diff
@@ -94,8 +89,6 @@ const EditorApp: React.FC<EditorAppProps> = ({
       manuscriptID,
       files,
       permittedActions,
-      userID: userID || '',
```
mbartenev-atypon, line 97: Maybe I'm missing something but it looks like userID is not getting to the store anymore. If so, can you please remove it from the state type and check if the app builds ok without it. (see src/store/Store.ts around line :39)

mbartenev-atypon (changes requested):

src/api/Api.ts
mnatsheh: I think we should display a "notification" of some kind to the user telling them that they need to login. Let's handle it in a separate PR.

mnatsheh (commented):
@mbetamony @mbartenev-atypon in my view, the entire TokenHandle logic (storage and expiration check) should live in the app (ops). From an article-editor perspective, it either receives a valid token, or nothing.

mbartenev-atypon:
> @mbetamony @mbartenev-atypon in my view, the entire TokenHandle logic (storage and expiration check) should live in the app (ops). From an article-editor perspective, it either receives a valid token, or nothing.

I think we can get rid of the tokenHandler indeed because it's for caching in the localStorage and we shouldn't handle it in the editor. In that case we can just call getAuthToken right in the axios interceptor. Alternatively we can pass the token down instead of a function but the editor may end up with a stale token and it will need either reloading or an action for user. So to pass a function is a more solid approach in my opinion.

We also need to notify dashboard team about this change to ensure they can adapt.

src/EditorApp.tsx
mbartenev-atypon: Please remove it from the store then. Thanks

src/api/Api.ts
mbetamony: One thing I'd like to ask here is, is it actually important to verify the token here? Shouldn't passing it down to the api be enough?

src/components/Page.tsx
mbetamony: Similar question to the discussion above

src/components/Page.tsx
```diff
@@ -47,13 +47,39 @@ const PageContainer = styled.div`
   font-family: ${(props) => props.theme.font.family.sans};
 `
 
-export const Page: React.FC<{ children: React.ReactNode }> = ({ children }) => {
-  const [{ tokenData }] = useStore((store) => ({
-    tokenData: store.tokenData,
-  }))
+export const Page: React.FC<{
```
mnatsheh, line 50: Is this actually needed?

mbetamony: I don't think so, passing it to the api should be enough i think (it won't load the document if the token is invalid anyways). That's what i was asking in the discussions above.

PR #817, opened 2025-05-04
Hide all editor body buttons that allow content editing while viewing snapshots - LEAN-4511

IslamJMomani:
Ensure editor remount when switching the editor mode.


src/components/projects/EditorElement.tsx
```diff
@@ -119,7 +120,11 @@ const EditorElement: React.FC = () => {
       <SpriteMap color="#353535" />
       {/* eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-static-element-interactions  */}
       <div id="editorDropzone" ref={drop}>
-        <div id="editor" ref={onRender}></div>
+        <div
+          id="editor"
+          key={`editor-mode-${isViewingMode ? 'view' : 'edit'}`}
```
mbartenev-atypon, line 125: are we doing this to force react recreate the html? Can you please add a comment in the code what it's for.

mbartenev-atypon (approved):

src/components/projects/EditorElement.tsx
IslamJMomani: Correct. Please let me know if the added comment makes the purpose clear.

mbartenev-atypon: Looks great. Thanks!

PR #841, opened 2025-05-20
LEAN-4533 SFI-010 - Compare Snapshots

fwaisi:


src/index.tsx
```diff
@@ -47,6 +47,7 @@ const ManuscriptEditor: React.FC<
   config,
   observer,
 }) => {
+  const params = useParams()
```
mbartenev-atypon, line 50: Looks like we dont need that

package.json
```diff
@@ -45,6 +45,7 @@
     "@types/use-sync-external-store": "^0.0.6",
     "autotrack": "^2.4.1",
     "axios": "^1.6.7",
+    "diff-match-patch": "^1.0.5",
```
mbartenev-atypon, line 48: I don't see that being used. Also there is no lock update if we need it.

src/components/projects/EditorElement.tsx
```diff
@@ -30,19 +30,21 @@ import { findParentNodeClosestToPos, flatten } from 'prosemirror-utils'
...
+import { ManuscriptPlaceholder } from './ManuscriptPlaceholder'
 
 const EditorElement: React.FC = () => {
   const [error, setError] = useState('')
   const [{ editor }] = useStore((store) => ({
     editor: store.editor,
   }))
 
-  const { onRender, view, dispatch } = useConnectEditor()
+  const { onRender, view, dispatch, isComparingDocuments } = useConnectEditor()
```
mbartenev-atypon, line 47: If you will prepare the document with comparison during Sources assembly stage as suggested in my comment in [src/hooks/use-create-editor.ts](https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/841/files#diff-5bf36822dcc2e4e9202ab1b425ff38501155b2e69cbf3dab773289427b2ff035) in this PR, you won't have to create any special loading procedures in the body editor. That also makes sense because the loading is already managed there and there is no need to make another loading as it's the only purpose of this instance - to compare snapshots.

src/hooks/use-create-editor.ts
```diff
@@ -49,6 +52,9 @@ export const useCreateEditor = () => {
     locale: store.cslLocale,
     sectionCategories: store.sectionCategories,
     isViewingMode: store.isViewingMode,
+    isComparingMode: store.isComparingMode,
```
mbartenev-atypon, line 55: here (a bit lower in this file) we send entire params object to body-editor while it just needs to know if it should enable comparison mode and as a result of that we also then need to ask body editor about wether it is in comparison mode. All this information should be established here, in the article editor, and body-editor doesn't need to know anything about the causes of enabling comparison mode - it should just receive a boolean "enableCompare" and the document with diff that is already prepared to be displayed by the article editor. Article editor shouldn't really just pass application level parameters to body-editor. Article editor is responsible for procuring snapshots and preparing documents so all that logics should be here. Preferably when you start the app, in the sources you can check for parameters and request 2 docs from api and create a new doc out of it. After that you will have identical flow as in the normal document like during readonly mode.

This approach should require little to no changes in body-editor and in the app in general, except for the logic related to managing interactive, and selection modal (which also should be in the article editor)

mbartenev-atypon (changes requested):

src/lib/compare-table-element.ts
```diff
@@ -0,0 +1,730 @@
...
+
+interface NodeWithAttrs extends ManuscriptNode {
+  attrs: Record<string, unknown>
+}
+
+/**
+ * Compare two table elements and track changes between them.
+ * This is the main entry point for table comparison.
+ */
+export const compareTableElement = (
```
mbartenev-atypon, line 32: I do really wonder if we need to create such deep comparison. Perhaps it would be enough just to say that the table is different. I'm not sure we can reasonably support diff of tables when extensive structural changes are made. We need to talk to Panos about it

src/components/inspector/SnapshotsList.tsx
```diff
@@ -10,22 +10,25 @@
...
 import styled from 'styled-components'
 
 import useExecCmd from '../../hooks/use-exec-cmd'
 import { ManuscriptSnapshot } from '../../lib/doc'
 import { useStore } from '../../store'
 import { FormattedDateTime } from '../FormattedDateTime'
+import { CompareDocumentsModal } from '../tools/CompareDocumentsModal'
 
 export const SnapshotsList: React.FC = () => {
+  const { id: submissionId } = useParams()
```
mbartenev-atypon, line 31: I'm not sure about this. The article-editor is not an app, it doesnt know anything bout pages and urls, so it shouldn't blindly access params from top level router. This should come as props from parent component that created the page.

src/lib/compare-documents.ts
```diff
@@ -0,0 +1,88 @@
+/*!
```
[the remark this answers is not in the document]
fwaisi: I have created a comparison dir and moved all related files.

src/lib/compare-table-element.ts
fwaisi: Yes, that’s a valid point.
But since the deep comparison logic is already implemented, and we’re delaying the page load until the comparison is ready, we can stick with it for now.

src/components/inspector/SnapshotsList.tsx
fwaisi: Yeah I agree with you, but for now the editor is only used in a route that always has the id, so it's not causing any issues. I think doing prop drilling or creating a new context just for this would be too much at this stage.

src/hooks/use-compare-documents.ts
```diff
@@ -0,0 +1,98 @@
...
+  isComparingMode: boolean
+  isLoading: boolean
+  error: string | null
+}
+
+export const useCompareDocuments = ({
+  originalId,
+  comparisonId,
+  getSnapshot,
+  originalDoc,
```
mbartenev-atypon, line 37: Not used

src/components/inspector/SnapshotsList.tsx
mbartenev-atypon: You can just put it into the store as everything else. Using params that essentially come out of nowhere is hardly a reliable approach.

src/lib/create-dataTracked-attrs.ts
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * All portions of the code written by Atypon Systems LLC are Copyright (c) 2025 Atypon Systems LLC. All Rights Reserved.
+ */
+import {
+  CHANGE_OPERATION,
+  CHANGE_STATUS,
+} from '@manuscripts/track-changes-plugin'
+import { v4 as uuidv4 } from 'uuid'
+
+export const createSetAttrsDataTracked = (
```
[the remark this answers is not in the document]
mbartenev-atypon: no comments or action provided on this

src/lib/distribute-nodes.ts
```diff
@@ -0,0 +1,185 @@
...
+          childOrderMap
+        )
+      })
+
+      // When processing the comparison document, reorder the map based on the order of nodes
+      if (!isOriginal) {
+        const orderedMap = new Map<string, NodeComparison>()
+
+        const sortedKeys = [...parentMap.keys()].sort((a, b) => {
+          const posA = orderMap.get(a) ?? 999
```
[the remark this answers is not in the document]
mbartenev-atypon: there is absolutely no guarantee that it's the limit of child nodes in any giving node. you need to use Number.MAX_VALUE. It will also help avoid seemingly meaningless numbers in your code

src/hooks/use-compare-documents.ts
```diff
@@ -0,0 +1,98 @@
...
+}: UseCompareDocumentsProps): UseCompareDocumentsReturn => {
```
mbartenev-atypon, line 38: I don't see a reason to specify return value type here. Typescript would derive the type out of the function return value. If that's just to convert Node into ManuscriptNode - you can just typecast that one variable in the return block of the function

src/lib/comparison/compare-documents.ts
```diff
@@ -0,0 +1,85 @@
...
+
+// The entry point for the comparison mode, which is used to compare two ManuscriptNode trees
+// It distributes the nodes, rebuilds the nodes tree and then creates a new ManuscriptNode tree
+export const compareDocuments = (
+  originalSnapshot: ManuscriptSnapshot,
+  comparisonSnapshot: ManuscriptSnapshot
+) => {
+  const originalDocument = schema.nodeFromJSON(originalSnapshot.snapshot)
+  const comparisonDocument = schema.nodeFromJSON(comparisonSnapshot.snapshot)
+  const originalTopLevelNodes = extractTopLevelNodes(originalDocument)
```
mbartenev-atypon, line 30: I don't see a point of pre-extracting top level nodes, you can easily start with a top level node already in distributeNodesForComparison since you are running a recursion there and avoid this operations.

src/lib/comparison/rebuild-nodes-tree.ts
```diff
@@ -0,0 +1,107 @@
...
+import { compareParagraphLike } from './compare-paragraph-like'
+import { compareTableElement } from './compare-table-element'
+import {
+  createDeleteAttrsDataTracked,
+  createInsertAttrsDataTracked,
+} from './create-dataTracked-attrs'
+import { NodeComparison } from './distribute-nodes'
+
+// Rebuilds the nodes tree from the distributed nodes, and applies the changes to the nodes
+export const rebuildDocNodeTree = (
```
mbartenev-atypon, line 24: what if you fine a node with the same id but different nodeType? (change node type type of change)

fwaisi: The currently supported node type changes are limited to converting a paragraph to a heading and vice versa.

Given how this feature is implemented, it will never result in a node having the same ID but a different node type.

src/lib/create-dataTracked-attrs.ts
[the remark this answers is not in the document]
fwaisi: I have added the typecast on the function return.

src/lib/rebuild-nodes-tree.ts
```diff
@@ -0,0 +1,108 @@
...
+      const childNode = rebuildDocNodeTree(childId, entry.children)
+      rebuiltChildren.push(childNode)
+    }
+  }
+  if (
+    entry.status === 'deleted' &&
+    !entry.comparisonNode?.isText &&
+    !entry.originalNode?.isText
+  ) {
+    console.log('deleted', entry.originalNode?.type.name)
```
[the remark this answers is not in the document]
fwaisi: fixed

src/lib/comparison/create-dataTracked-attrs.ts
```diff
@@ -0,0 +1,68 @@
...
+ * All portions of the code written by Atypon Systems LLC are Copyright (c) 2025 Atypon Systems LLC. All Rights Reserved.
+ */
+import {
+  CHANGE_OPERATION,
+  CHANGE_STATUS,
+} from '@manuscripts/track-changes-plugin'
+import { v4 as uuidv4 } from 'uuid'
+
+// Creates a dataTracked attribute for a set node attributes operation
+export const createSetAttrsDataTracked = (
```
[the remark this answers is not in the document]
fwaisi: I appreciate you pointing that out again—some comments were unfortunately missed due to the extent of recent changes. I’ve updated the return value with the appropriate typecast and revised the function names to better match their comparison-related purpose.

src/lib/rebuild-nodes-tree.ts
```diff
@@ -0,0 +1,108 @@
...
+    return baseNode.type.create(
+      finalAttrs,
+      rebuiltChildren.length > 0 ? rebuiltChildren : baseNode.content
+    )
+  } else if (
+    entry.comparisonNode?.type == schema.nodes.table_element &&
+    entry.originalNode?.type == schema.nodes.table_element
+  ) {
+    // TODO: find more generic way to handle this
+    return compareTableElement(entry.originalNode, entry.comparisonNode)
```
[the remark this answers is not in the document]
fwaisi: Yes, but showing changes separately gives users a clearer and more accurate comparison.

src/lib/rebuild-nodes-tree.ts
```diff
@@ -0,0 +1,108 @@
...
+  } else if (entry.comparisonNode?.isBlock && entry.originalNode?.isBlock) {
+    return compareSingleNodeAttrs(
+      entry.originalNode,
+      entry.comparisonNode,
+      baseNode.type,
+      rebuiltChildren
+    )
+  }
+
+  const finalAttrs = { ...baseNode.attrs }
```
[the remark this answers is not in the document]
fwaisi: fixed

src/lib/comparison/create-dataTracked-attrs.ts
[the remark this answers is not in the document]
mbartenev-atypon: No worries. Thanks for taking care of that!

mbartenev-atypon (approved):

src/components/inspector/SnapshotsList.tsx
fwaisi: Actually I have address it, check the following https://github.com/Atypon-OpenSource/manuscripts-article-editor/pull/841/commits/ac728840b623fff46ee278148952eb2d0b4d93ac#diff-2b3973aba0936b36254b5107ae76dcf6dc5e71be8745a871daaaaf90969c01a6

I think you are checking outdated file changes.

src/hooks/use-create-editor.ts
```diff
@@ -49,35 +51,51 @@ export const useCreateEditor = () => {
     locale: store.cslLocale,
     sectionCategories: store.sectionCategories,
     isViewingMode: store.isViewingMode,
+    getSnapshot: store.getSnapshot,
   }))
 
   const api = useApi()
+  const params = useParams()
```
mbartenev-atypon, line 58: what I wrote about use params on submissionId also applies here. Thanks!

mbartenev-atypon: as per IM, we can do it later, this PR becomes a bit too much and the code is ok to be merged as it is now.

src/components/inspector/SnapshotsList.tsx
mbartenev-atypon: My bad, I mixed it up with the other useParams use case

PR #860, opened 2025-06-04
MPF-015 - Support for multiple images in a figure

nghazzawi-atypon:


src/components/FileManager/FileActions.tsx
```diff
@@ -58,6 +60,7 @@ export const FileActions: React.FC<{
   const showDownload = can?.downloadFiles && onDownload
   const showReplace = can?.replaceFile && onReplace
   const showDetach = can?.detachFile && onDetach
+  const showDelete = can?.editArticle && onDelete
```
mbartenev-atypon, line 63: hmm... maybe we need to can.detachFile here since its the closest in meaning? @mnatsheh

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -49,45 +61,39 @@ export const InlineFilesSection: React.FC<InlineFilesSectionProps> = ({
...
         icon = FileFigureIcon
       }
       return {
         element,
-        node: figure?.node,
-        pos: figure?.pos,
-        file,
         label,
         icon,
+        files: element.files.map((f) => f.file || { id: '', name: '' }),
```
mbartenev-atypon, line 94: so if there is no file, will it show up as empty field? How will "{ id: '', name: '' }" exactly work in this case?

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -96,58 +102,135 @@ export const InlineFilesSection: React.FC<InlineFilesSectionProps> = ({
...
     tr.scrollIntoView()
     view.focus()
     view.dispatch(tr)
   }
 
   return (
     <>
-      {metadata.map((e, index) => {
+      {groupedMetadata.map((group, groupIndex) => {
+        const [isOpen, setOpen] = useState(true)
```
mnatsheh, line 165: I'm not sure this is valid. We should move the state to the top level and maybe store a list of open nodes instead.

src/components/FileManager/FileActions.tsx
mnatsheh: Agreed.

src/components/FileManager/InlineFilesSection.tsx
nghazzawi-atypon: @mbartenev-atypon  My intention with using the fallback { id: '', name: '' } was to explicitly represent elements that should have an associated file but currently do not have one linked (please check the screenshot), this results in placeholders (like "Figure") being displayed in the "Inline files"  for such entries. This approach allowing users to see that a content element (e.g., a figure ) exists even if its file is missing
Alternatively, we could implement a solution where entries with missing f.file are filtered out entirely from the groupedMetadata 
. This would mean such elements would not appear in the "Inline files" list at all.

Screenshot from 2025-06-11 15-42-45

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -96,58 +102,135 @@ export const InlineFilesSection: React.FC<InlineFilesSectionProps> = ({
...
-            <FileActions
-              data-cy="file-actions"
-              sectionType={FileSectionType.Inline}
-              onReplace={async (f) => await handleReplace(e, f)}
-              onDetach={() => handleDetach(e)}
-              onDownload={() => fileManagement.download(e.file)}
-            />
-          </FileContainer>
+            <FileGroupHeader>
+              <div>
```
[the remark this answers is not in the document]
nghazzawi-atypon: Actually, it was necessary to add a flex container for the figure element's icon and label, and also to apply display: flex to the parent container. This is because we added a collapse button to the right of the figure label and icon, so yes, this change is needed.

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -14,30 +14,42 @@ import {
...
 type FileMetadata = {
   element: ElementFiles
   label: string
   icon: React.FC<React.SVGAttributes<SVGElement>>
-  file: FileAttachment
-  node?: ManuscriptNode
-  pos?: number
+  files: FileAttachment[]
+  nodes?: ManuscriptNode[]
+  positions?: number[]
```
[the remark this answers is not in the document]
nghazzawi-atypon: done

src/components/FileManager/FileActions.tsx
nghazzawi-atypon: done

src/components/FileManager/InlineFilesSection.tsx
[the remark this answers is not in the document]
mnatsheh: why do we need an intermediate container for display flex to work? Maybe I'm missing something. @IslamJMomani can you follow up with Noura on this?

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -14,30 +14,40 @@ import {
...
+import { FileTypeIcon } from './FileTypeIcon'
 
 type FileMetadata = {
   element: ElementFiles
   label: string
   icon: React.FC<React.SVGAttributes<SVGElement>>
-  file: FileAttachment
-  node?: ManuscriptNode
-  pos?: number
+  files: { file: FileAttachment | null, pos: number }[]
```
mnatsheh, line 46: Can we use `NodeFile[]` here? it looks like we're getting the value from `element.files` which has the type `NodeFile[]`.

src/components/FileManager/InlineFilesSection.tsx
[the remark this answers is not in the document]
nghazzawi-atypon: @mnatsheh 
we don't need an intermediate container for display flex , The goal was to achieve a specific layout: the icon and the figure label (Figure 1:) grouped together on the left, and the collapse/expand toggle button (ToggleIcon) positioned independently on the far right.

The justify-content: space-between; property on FileGroupHeader is designed to distribute space between its direct flex children, pushing the first item to the start and the last item to the end, and spreading any items in between.

If the icon, label, and toggle button were all direct children of FileGroupHeader, justify-content: space-between; would try to put space between the icon and the label, and then between the label and the toggle button, effectively spreading all three elements out. This would break the visual requirement for the icon and label to stick together.

Therefore, the div was originally added to serve as a single logical flex item for FileGroupHeader's justify-content: space-between;. This div wrapped both the icon and the label, making them behave as one unit.

**However, to achieve a  more direct control, I have removed that intermediate div and updated the fix by Refined CSS for FileGroupHeade and simplify the component *

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -49,105 +59,201 @@ export const InlineFilesSection: React.FC<InlineFilesSectionProps> = ({
...
+                      handleFileClick(fileAttachment.pos)
+                    }}
+                  >
+                    {fileAttachment.file && (
+                      <FileTypeIcon file={fileAttachment.file} />
+                    )}
+                    <FileNameText data-cy="filename">
+                      {fileAttachment.file?.name
+                        ? trimFilename(fileAttachment.file.name, 25)
+                        : 'Figure'}
```
mnatsheh, line 222: I don't think 'Figure' is appropriate when the file doesn't exist. I suggest something like 'Unknown file'.

src/components/FileManager/InlineFilesSection.tsx
```diff
@@ -49,105 +59,201 @@ export const InlineFilesSection: React.FC<InlineFilesSectionProps> = ({
...
+                      <Tooltip
+                        id={`${fileAttachment.file?.id}-file-name-tooltip`}
+                        place="bottom"
+                      >
+                        {fileAttachment.file?.name || 'Figure'}
```
mnatsheh, line 227: We don't need a tooltip if the file doesn't exist.

src/components/FileManager/FileGroup.tsx
```diff
@@ -0,0 +1,109 @@
...
+  cursor: pointer;
+  padding-right: 20px;
+
+  .file-icon {
+    margin-right: ${(props) =>
+      props.theme.grid.unit * 2}px; /* Adjust as needed */
+  }
+
+  ${FileLabel} {
+    margin-right: auto;
```
mnatsheh, line 74: You don't need a margin for this. You can use something like `flex: 1` to force an element to take all the empty space.

mnatsheh (approved):

mnatsheh (approved):

PR #862, opened 2025-06-10
LEAN-4636 - add editor state hook

mnatsheh:


src/hooks/use-editor-state.ts
```diff
@@ -0,0 +1,39 @@
...
+import { isEqual } from 'lodash'
+import { useCallback, useContext, useEffect, useState } from 'react'
+
+import { state } from '../store'
+import { AppStateObserverContext } from './use-app-state-observer'
+
+export const useEditorState = <T>(selector: (s: state) => T) => {
+  const [value, setValue] = useState<T>()
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  selector = useCallback(selector, [])
```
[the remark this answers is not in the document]
mbartenev-atypon: This would be hard to implement in a reasonable way, since you wont be able to adjust selector once component is mounted in the hosting app. I know it's just copied but it was transparent when implemented directly in the hosting app and now it isn't

src/EditorApp.tsx
```diff
@@ -10,20 +10,15 @@
  * All portions of the code written by Atypon Systems LLC are Copyright (c) 2019 Atypon Systems LLC. All Rights Reserved.
```
mbartenev-atypon, line 10: Why aren't we removing original implementation of the observer?

src/hooks/use-app-state-observer.ts
```diff
@@ -0,0 +1,59 @@
+/*!
```
mbartenev-atypon, line 1: can we please move those hooks to some special folder that would make it clear that these hooks are not for internal use? Something like externalHooks or clientHooks or consumerHooks? Would be great also if we can annotate them to make it clear why they are supplied since it's not really a part of article editor but some helping tools for hosting apps.

src/EditorApp.tsx
mnatsheh: Which one is that?

src/hooks/use-editor-state.ts
```diff
@@ -0,0 +1,39 @@
...
+  const observer = useContext(AppStateObserverContext)
+  useEffect(() => {
+    if (!observer) {
+      return
+    }
+    const sub = (state: state) => {
+      if (state) {
+        const $new = selector(state)
+        setValue((old) => (isEqual(old, $new) ? old : $new))
```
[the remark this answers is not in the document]
mnatsheh: Replaced the lodash implementation with the deeper-equal implementation.

src/hooks/use-app-state-observer.ts
mnatsheh: Moved the files (now a single file) to hooks/external.

src/hooks/use-editor-state.ts
```diff
@@ -0,0 +1,39 @@
...
+      }
+    }
+    observer.subscribe(sub)
+    return () => {
+      observer.unsubscribe(sub)
+    }
+  }, [observer, selector])
+  return [value]
```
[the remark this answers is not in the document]
mnatsheh: I think the rational I had for returning an array here is to match the useState signature. I can return the value directly if you prefer.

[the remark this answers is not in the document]
mbartenev-atypon: hm, I see. Not sure if we need to match it but it's ok if that looks more appropriate to you.

src/EditorApp.tsx
mbartenev-atypon: It look to me like you integrated article-editor's side of observer implementation into it. I see now it's not the case. Sorry about the confusion.

PR #866, opened 2025-06-15
LEAN-4614 NM-030 - Notifications for Referenced Track changes

fwaisi:


src/components/inspector/IssuesPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+          <NoIssuesMessage>No issues found</NoIssuesMessage>
+        ) : (
+          warnings.map((warning: Warning, index: number) => (
+            <WarningItem
+              key={index}
+              isFocused={selectedWarningIndex === index}
+              onClick={() => handleWarningClick(warning, index)}
+              data-cy="warning"
+            >
+              <WarningMessage>{warning.message}</WarningMessage>
```
mbartenev-atypon, line 159: generally in this implementation warning and inconsistency used interchangeably. We really should pick just one term an used it for everything. Inconsistency sounds better to me since it reflects the meaning in a more precise way than warning.

src/components/inspector/IssuesPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+        </CollapseIcon>
+      </CollapsibleHeader>
+
+      <WarningsList isVisible={!isCollapsed}>
+        {warnings.length === 0 ? (
+          <NoIssuesMessage>No issues found</NoIssuesMessage>
+        ) : (
+          warnings.map((warning: Warning, index: number) => (
+            <WarningItem
+              key={index}
```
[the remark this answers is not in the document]
mbartenev-atypon: Probably worth checking

src/components/inspector/IssuesPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+  }
+
+  svg {
+    width: 19px;
+    height: 19px;
+  }
+`
+
+const WarningsList = styled.div<{ isVisible: boolean }>`
+  display: ${(props) => (props.isVisible ? 'block' : 'none')};
```
mbartenev-atypon, line 70: I wonder why we are hiding it with css and not just unmount it with react i.e.` {isVisible && <WarningsList />}`

src/hooks/use-connect-editor.ts
```diff
@@ -66,6 +70,14 @@ export const useConnectEditor = () => {
     storeDispatch({ hasPendingSuggestions })
   }, [storeDispatch, hasPendingSuggestions])
 
+  const warningsCount = useMemo(() => {
+    return detectInconsistencyPluginKey.getState(state)?.warnings?.length || 0
+  }, [state])
+
+  useEffect(() => {
+    storeDispatch({ warningsCount })
```
mbartenev-atypon, line 78: would that be worth just putting all the warning in the store instead of just count? they are going to be used the inspector anyway. What do you think?

src/components/inspector/IssuesPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+`
+
+const WarningItem = styled.div<{ isFocused: boolean }>`
```
mbartenev-atypon, line 73: It seemed to me in the designs. That warning items is sharing the styles with a suggestion item. Can we reuse the styling from there?

mbartenev-atypon (changes requested):

src/lib/change-handlers.ts
```diff
@@ -194,10 +194,11 @@ export const handleNodeChange = (
         content: node.attrs.source,
       }
     }
+    // TODO: checking why ts is not complaining about the type of node.attrs.type as parameter in getTitleDisplayName
```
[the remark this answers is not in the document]
fwaisi: done

src/components/inspector/IssuesPanel.tsx
```diff
@@ -0,0 +1,166 @@
...
+`
+
+const HeaderTitle = styled.h2`
+  margin: 0;
+  font-size: ${(props) => props.theme.font.size.large};
+  font-weight: 600;
+  color: ${(props) => props.theme.colors.text.primary};
+`
+
+const CollapseIcon = styled.button<{ isCollapsed: boolean }>`
```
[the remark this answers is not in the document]
fwaisi: Actually its interactive which rotate based on the collapse state.
I will change it to span, instead of button.

src/components/inspector/IssuesPanel.tsx
fwaisi: done

src/components/inspector/IssuesPanel.tsx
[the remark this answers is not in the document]
mbartenev-atypon: Right, but the parent is button anyway

src/components/inspector/IssuesPanel.tsx
mbartenev-atypon: Ok, please file that ticket

PR #894, opened 2025-07-03
MFP-020 - Reorder pictures within a figure

nghazzawi-atypon:
Using context={window} to prevent "Cannot have two HTML5 backends at the same time" error.
     * This allows components rendered outside the main React tree (like modals via ReactSubView)
     * to access the same DndProvider context, avoiding conflicts when multiple React roots
     * try to initialize their own HTML5Backend instances.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -21,8 +21,8 @@ importers:
         specifier: 2.2.3
         version: 2.2.3(react-dom@18.3.1(react@18.3.1))(react@18.3.1)
       '@manuscripts/body-editor':
-        specifier: 3.2.29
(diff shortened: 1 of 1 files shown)


src/Main.tsx
```diff
@@ -26,7 +26,9 @@ const Main: React.FC<EditorAppProps> = ({
   getAuthToken,
   observer,
 }) => (
-  <DndProvider backend={HTML5Backend}>
+  <DndProvider backend={HTML5Backend} context={window}>
```
fwaisi, line 29: I believe this issue only occur locally in development mode especially with linking between repos

PR #939, opened 2025-08-18
Search for a text in read-only mode - LEAN-4809

IslamJMomani:
Enable global keyboard shortcuts. Useful for viewer mode where users can't focus the editor but still need shortcuts, and when users are elsewhere on the page (inspector).

Note: 

- In this PR I added search-related shortcuts (Cmd+f, Shift+Cmd+h) as global shortcuts because:

    1. This addresses the specific ticket scope
    2. Search operations are selection-independent

-  I intentionally didn't include other editing shortcuts because they require active selection context. Activating these while the editor isn't focused would cause unexpected behavior.

- We can incrementally add other selection-independent actions as global shortcuts when needed.

@mbartenev-atypon, Please take a look.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@manuscripts/article-editor",
-  "version": "4.4.7",
+  "version": "4.4.8",
   "license": "CPAL-1.0",
   "description": "React components for editing and viewing manuscripts",
   "repository": "github:Atypon-OpenSource/manuscripts-article-editor",


src/hooks/use-global-keyboard-shortcuts.ts
```diff
@@ -0,0 +1,61 @@
...
+export const useGlobalKeyboardShortcuts = () => {
+  const [editor] = useStore((state) => state.editor)
+
+  useEffect(() => {
+    if (!editor?.view) {
+      return
+    }
+
+    const { view } = editor
+    const keymap = {
```
mbartenev-atypon, line 43: just a general note on using object, I would prefer for us to move towards using Maps instead of plain objects since they don require any trickery such as Object.entries, Object.keys etc, and have a better typing support. But that is of course optional.

IslamJMomani: Good suggestion! I refactored to use Map. Thanks for the tip!

PR #956, opened 2025-08-28
LW-FY26-Q2: The Files full name doesn't display fully (LEAN-4887)

asouqi:
The issue with this ticket file names appears as they have the same name. 
image-2025-08-28-01-47-12-717

will use [middle truncate](https://ni.github.io/design-system/patterns/truncation/), as most UI design guidelines recommend that


src/components/FileManager/FileName.tsx
```diff
@@ -23,10 +23,7 @@ export const FileName: React.FC<{
   label?: string
   maxBaseNameLength?: number
 }> = ({ file, label, icon: Icon, maxBaseNameLength = 25 }) => {
-  const adjustedMaxBaseNameLength = label
-    ? maxBaseNameLength - label.length
-    : maxBaseNameLength
-  const trimmedFilename = trimFilename(file.name, adjustedMaxBaseNameLength)
+  const trimmedFilename = trimFilename(file.name, maxBaseNameLength)
```
asouqi, line 26: @fwaisi QA assumes the max length to be the same for all tapes, if that is not expected can you please add comments in LEAN-4887 there that trim will not be the same for all tapes

asouqi: nevermind I checked that with Panos

fwaisi: Noted.

PR #973, opened 2025-09-14
Track supplements changes - LEAN-4810

IslamJMomani:


src/lib/change-handlers.ts
```diff
@@ -145,7 +146,8 @@ export const handleTextChange = (
 
 export const handleNodeChange = (
   suggestion: NodeChange | NodeAttrChange,
-  state: ManuscriptEditorState
+  state: ManuscriptEditorState,
+  getFiles?: FileAttachment[]
```
mbartenev-atypon, line 150: the name suggest a function, but you receive an array. Please consider renaming to just "files" or similar.

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
```diff
@@ -74,7 +74,7 @@ export const SuggestionSnippet: React.FC<Props> = ({
         ChangeSet.isNodeChange(suggestion) ||
         ChangeSet.isNodeAttrChange(suggestion)
       ) {
-        newSnippet = handleNodeChange(suggestion, view.state)
+        newSnippet = handleNodeChange(suggestion, view.state, getState().files)
```
mbartenev-atypon, line 77: hm, why not just pass files right from the store without getState() and introduce it as a dependency on the useEffect?

mbartenev-atypon (changes requested):

src/lib/change-handlers.ts
IslamJMomani: Done

src/components/track-changes/suggestion-list/SuggestionSnippet.tsx
IslamJMomani: No reason 🤦‍♀️

mbartenev-atypon (approved):

PR #987, opened 2025-09-29
CI-210 - Orphaned comments moved to a special section (LEAN-4818)

asouqi:
[Design](https://www.figma.com/design/7W1KAHVqHnHqYtxk0GKsEc/Designs-for-orphan-comments?node-id=2648-53934&p=f&m=dev)
- will show comment as orphan when the delete change gets approved
- in the design it will not show delete button at the orphan comments reply, as deleting a comment will do a clean up to it's related replies also

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@manuscripts/article-editor",
-  "version": "4.6.17",
+  "version": "4.6.18",
   "license": "CPAL-1.0",
   "description": "React components for editing and viewing manuscripts",
   "repository": "github:Atypon-OpenSource/manuscripts-article-editor",


src/lib/comments.ts
```diff
@@ -41,6 +44,39 @@ export const buildThreads = (
     }))
 }
 
+/**
+ *  will filter out orphan comments with their reply nodes from comments plugin,
```
IslamJMomani, line 48: The comment is a bit misleading, it says `filter out orphan comments from comments plugin` but actually the comments coming from the plugin already filtered and does not include orphan comments.

asouqi: updated

PR #1156, opened 2026-05-05
LEAN-5550 User Info from Parent App

mbartenev-atypon:


src/components/projects/EditorHeaderGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+                toggleTrackingChangesVisibility()
+              }
+            }}
+            tabIndex={0}
+            role="button"
+            aria-label="show tracked changes"
+            aria-pressed={isTrackingChangesVisible}
+            data-tooltip-content={compact ? label : ''}
+          >
+            {!compact && <Label>Show tracked changes</Label>}
```
asouqi: can't we do that with css @media

src/components/projects/EditorHeaderGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+              if (e.key === 'Enter') {
+                e.preventDefault()
+                toggleTrackingChangesVisibility()
+              }
+            }}
+            tabIndex={0}
+            role="button"
+            aria-label="show tracked changes"
+            aria-pressed={isTrackingChangesVisible}
+            data-tooltip-content={compact ? label : ''}
```
asouqi: This component is a div and `data-tooltip-content` is a custom attribute as far as I know, so that will not have any effect

src/components/projects/EditorHeaderGroup.tsx
```diff
@@ -0,0 +1,151 @@
...
+  const [{ savingProcess, isViewingMode }] = useStore((store) => ({
+    isViewingMode: store.isViewingMode,
+    isComparingMode: store.isComparingMode,
+    savingProcess: store.savingProcess,
+  }))
+
+  const container = useRef<HTMLDivElement>(null)
+  const [compact, setCompact] = useState(false)
+
+  useLayoutEffect(() => {
```
asouqi: should't we have a design for each screen size, and will control each one with css mostly, instead of listening to screen resize changes?

asouqi (changes requested):

src/components/projects/EditorHeaderGroup.tsx
mbartenev-atypon: Well, we don't have any designs except for a large desktop, unforutnately. And this is not a design based work but just a fix for smaller windows or devices that the designs dont cover. The reason why I opted for the in-code solution is because I initially developed it for the the LEAN-5459, where there was a roster of users represented by round avatars that need that info anyway and, because it took up some space in the same line as the track-changes switch, the designs broke on much larger screens. In this implementation here the "compact" boolean is necessary only for the data-tooltip. The label I can hide with the query but not sure if it's worth it, since the compact is already there. Let me know what you think please.

src/components/projects/EditorHeaderGroup.tsx
mbartenev-atypon: data-tooltip is functional globally - you dont need a special component for it. I tried it before submitting and it worked.

src/store/Store.ts
```diff
@@ -38,6 +38,13 @@ export type action = { action?: string; [key: string]: any }
 
 export type PMEditor = ReturnType<typeof useCreateEditor>
 
+export type HostUser = {
+  id: string,
+  displayName: string,
+  firstName?: string
+  lastName?: string
```
asouqi, line 47: we need to add connectId

asouqi (changes requested):

src/components/projects/EditorHeaderGroup.tsx
asouqi: what about using `data-tooltip-class-name` to decide if we can hide/show tooltip, like this:
```
import { createGlobalStyle } from 'styled-components';

const TooltipStyles = createGlobalStyle`
  @media (min-width: 651px) {
    .hide-tooltip-on-desktop {
      display: none !important;
    }
  }
`;


 <TrackChangesToggleWrapper
            data-tooltip-content={label}
            data-tooltip-class-name={"hide-tooltip-on-desktop"}
          >
```

https://github.com/user-attachments/assets/e4a8a293-c933-4c21-83a9-c4ee3b98dd33

asouqi (approved):
I don't mind the change in EditorHeaderGroup for screen size, but I prefer to keep that in CSS. I will approve the PR to avoid further delays

asouqi:
@mbartenev-atypon please don't forget to pass users props in both:

https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/Main.tsx
https://github.com/Atypon-OpenSource/manuscripts-article-editor/blob/master/src/index.tsx

src/components/projects/EditorHeaderGroup.tsx
mbartenev-atypon: Thanks for the tip! Updated to user css to control the display.

PR #1159, opened 2026-05-06
LEAN-5661

fwaisi:


mnatsheh (changes requested):
There shouldn't be any custom logic in FE for this case. If any logic is needed, it should be under wm-hooks.

fwaisi:
> There shouldn't be any custom logic in FE for this case. If any logic is needed, it should be under wm-hooks.

Done

mnatsheh:
Why do we need this still?

fwaisi:
> Why do we need this still?

I was handling the error separately, but now it’s included in useFileManagement.

The remaining commits are only for removing the client-side validation.

mnatsheh (approved):

PR #1167, opened 2026-05-11
remove sorted snapshots

mbetamony:
Based on the ticket, they want to reverse the ordering of the snapshots. Maybe we should rely on backend to return the order of the snapshots.


mnatsheh (approved):
If what you're saying is that the api returns the snapshots already sorted, then yes we should just depend on that.

PR #1193, opened 2026-06-17
LEAN-3299: Update styled-components

AlaAlkhalaileh:


src/components/comments/CommentCard.tsx
```diff
@@ -48,7 +48,7 @@ const Timestamp = styled(RelativeDate)`
   white-space: nowrap;
   padding-right: 8px;
 `
-const RepliesCount = styled.div.attrs({
+const RepliesCount = styled.div.attrs<{ 'data-cy': string }>({
```
mbartenev-atypon, line 51: hm, are you sure this needs to be specified? Shouldn't data- prefixed attributed be supported without explicit typing?

src/components/projects/ManuscriptToolbar.tsx
```diff
@@ -30,7 +30,7 @@ export const ToolbarItem = styled.div`
   position: relative;
 `
 
-const ToolbarButton = styled.button.attrs({
+const ToolbarButton = styled.button.attrs<{ 'data-toolbar-button': boolean }>({
```
mbartenev-atypon, line 33: Same here

src/components/comments/CommentCard.tsx
AlaAlkhalaileh: I don't know why, but looks like styled-component dev team decided to make styled-components v6's .attrs() type is strict and doesn't allow arbitrary data-* attributes without the explicit type parameter.

image

mbartenev-atypon (approved):

